PR #22, opened 2017-06-13
feat(license): Add CC-BY-SA-4.0 License

raisedadead:
Add some license goodies.

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,0 +1,427 @@
+Attribution-ShareAlike 4.0 International
+
+=======================================================================
+
+Creative Commons Corporation ("Creative Commons") is not a law firm and
+does not provide legal services or legal advice. Distribution of
+Creative Commons public licenses does not create a lawyer-client or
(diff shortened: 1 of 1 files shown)


Bouncey:
@QuincyLarson is this the correct licence?

raisedadead:
This would be for the content and maybe we can add a dual license for the software, as BSD-3. But this should be the one the GitHub shows in the tab because of the primary content of the repo.

Bouncey:
Ah, the content is going to be hosted in a different repo that is linked to [Netlify CMS](https://www.netlifycms.org/). The content repo does not exist yet though.

A BSD-3 for this repo would be great though. :+1:

QuincyLarson:
@raisedadead we're going to store the code for the guide in this repo, and the guide articles themselves in https://github.com/freeCodeCamp/guide-articles/ so that would be the repo where we'd want to use the CC license. In this repo, we'd use the BSD-3 license.

PR #27, opened 2017-06-24
Navbar

kuriakinzeng:
The navbar is in a working state. As mentioned:
1. I have made clicking on a title on side menu = clicking on the caret = open the dropdown submenu. If we move forward with this pattern, we need to list the index/introduction for each topic listed as the first item in the dropdown submenu
2. I have ditched React-Bootstrap. This code requires two navbars (one TopNavBar and one SideNavBar) where the toggle button on TopNavBar controls SideNavBar and Content component. The whole point of using React-Bootstrap is that it could handle menu collapsing via Navbar.Toggle and Navbar.Collapse, but I found that the two have to be in wrapped in the same Navbar for them to work properly.


Bouncey:
Re: `react-bootstrap` I thought the idea was an onscreen menu whilst on desktop and a toggled nav on mobile. We can use react to conditionally render based on screen size.

I would have a preference for `react-bootstrap` as it reduces clutter in our code base.

Bouncey:
with the added bonus of not requiring jQuery

kuriakinzeng:
Well that's another way of doing it, but I'm not sure if it's worth the time and delay because:
1. This code doesn't use jQuery. It's fully done with React and Bootstrap.css
2. I intend to outsource the SideBarNav and Content in `pages/_template.jsx` into its own component. It's not something that you have to look at ever again since it will be independent. 
3. The css needs to be there even if we use React-Bootstrap. I'm not convinced that using React-Bootstrap will result in less code, since it will be almost one-to-one replacement of native bootstrap tags with React-Bootstap's component which is just a wrapper of the native bootstrap code.

Bouncey:
I think I would still like to see `react-bootstrap` implemented so we can use a bunch of nested accordions for the drill down menu, which would be easy to control with react

If you are occupied with writing content then I would be happy to take this up :+1:

QuincyLarson:
@kuriakinzeng I agree with @Bouncey that we need multi-nested sidenav functionality. If react-bootstrap is necessary, perhaps we could find an actively maintained fork that uses Bootstrap 3.0?

Bouncey:
@QuincyLarson `react-bootstrap` uses bootstrap v3.x

kuriakinzeng:
I can do it but it will not be until the week after. It would be great if @Bouncey can work on it. Thanks @Bouncey and @QuincyLarson

QuincyLarson:
@kuriakinzeng OK - sounds good.

@Bouncey do you have time to tackle this? I'm happy to help QA and give feedback.

Bouncey:
I will close this in favour of `react-bootstrap` which offers more 'fine grained' control over the UI. Thanks for the work you put in to this @kuriakinzeng, I can use the layout you proposed 👍 

@QuincyLarson I will aim to have the first iteration in a PR through the weekend

PR #28, opened 2017-07-02
Bootstrap the UI

### LINKED ISSUE #5: Get nav bar logo to look like the logo on  https://www.freecodecamp.com/donate/ with similar size logo and placement.

Bouncey:
closes #5

https://youtu.be/rFa0vbS9Jug

In the video you will see an issue I am having that I cannot get my head around.

Clicking on the title of an article does not navigate to that article, although the link is there as shown by the copy and paste of the link into the address bar. 

You can see when I click a clojure article that the url is updated and reverted almost instantly. 

This is the code for the behaviour of the article title 

```js
import React from 'react';
import PropTypes from 'prop-types';

const propTypes = {
  path: PropTypes.string,
  title: PropTypes.string
};

class NavItem extends React.PureComponent {
  constructor() {
    super();

    this.handleClick = this.handleClick.bind(this);
  }


  handleClick() {
    const { path } = this.props;
    this.context.router.push(path);
  }

  render() {
    const { path, title } = this.props;
    return (
      <li onClick={ this.handleClick }role='presentation'>
        <span>{ title }</span>
      </li>
    );
  }
}

NavItem.displayName = 'NavItem';
NavItem.propTypes = propTypes;
NavItem.contextTypes = {
  router: React.PropTypes.object.isRequired
};

export default NavItem;
```

I have tried navigation with `this.context.router` and with `this.props.router` which becomes available through the `withRouter` HOC.

I have tried the `<Link>` from `react-router` 
```sh
Uncaught Error: Attempted to update component `NavItem` that has already been unmounted (or failed to mount).
```
and `<LinkContainer>` from `react-router-bootstrap`.
```sh
Warning: Failed context type: The context `router.history` is marked as required in `LinkContainer`, but its value is `undefined`.
```

The fact that I have tried all these different things is telling me that I have messed something up, but I am blind to it.

@BerkeleyTrue if you can see where I am going wrong I would greatly appreciated it.


Bouncey:
@BerkeleyTrue I fixed it now!!

PR #35, opened 2017-07-08
Fix linting issues

Bouncey:
This PR addresses linting issues that are making all other PR's fail Travis-CI


Bouncey:
The remaining failing tests are due to the node version be being tested. 

Node versions 4+ are passing. 

#31 addresses this

PR #51, opened 2017-07-13
Reduce Travis-CI workload

### LINKED ISSUE #50: Optimize Travis CI
The CI builds are taking on an average of 30 mins. 
This is creating a trickle-down effect on all other repositories under the org as well.

Do we really need so many config types i.e: 2 OS-types X 3 Node Env = 6 separate configs?
Considering these are C++ builds it's adding even more to the workload on the builds.

Bouncey:
closes #50

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,9 +2,7 @@
 language: cpp
 sudo: false
 env:
-  - export NODE_VERSION="4"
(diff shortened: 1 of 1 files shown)


Bouncey:
Travis passed on Linux

PR #71, opened 2017-07-18
Add sql create table statement article

vanessasena:


QuincyLarson:
@vanessasena Excellent work. This is a nice quick introduction to how SQL Create works.

PR #75, opened 2017-07-19
Fix the link to the "guides website"

nsuchy:
This change fixes the link to the guide website.


Bouncey:
Good catch :+1: 

Thanks @nsuchy

PR #78, opened 2017-07-20
Removed duplicate page.

nsuchy:
This commit removes a duplicate page, there does not need to be a Bootstrap Buttons CSS reference as the CSS for Bootstrap Buttons is clearly explained on the Bootstrap Buttons page.


Bouncey:
@nsuchy Unfortunately, simply deleting the `index.md` for an article will not completely remove the page from the guide site. We also need to remove the containing folder to stop gatsby looking for a page that isn't there.

Feel free to raise a PR that removes the containing folder also

Many thanks and happy coding :sparkles:

PR #84, opened 2017-07-21
Added .DS_Store files to Git Ignore

nsuchy:
This pull request adds .DS_Store files to Git Ignore, developers on macOS may accidentally commit these local files to the Git repo - it would make sense to add it to Git Ignore to accommodate for those developers. If the maintainers agree with this change, it should be merged.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,8 @@
+# Custom Ignores
 package-lock.json
-
 navData.json
 .svn
+*.DS_Store
 
 # Logs
 logs


jimtryon:
I agree, this would be useful for those who develop in Mac environments.

.gitignore
```diff
@@ -34,3 +34,7 @@ node_modules
 .sass-cache/
 public/
 .cache/
+
+# .DS_Store Files
```
Bouncey: Ok, so this is an error I made. Yes I agree with addin `.DS_Store` to the `.gitignore` file, but I had an area set aside for custom ignores, I just didn't mark it out correctly.

Lines 1-4 are our custom ignores for this repo, the rest were added by `Gatsby`

Can you add a comment at the top of the file donating custom ignores and move `.DS_Store` there please

Many thanks

nsuchy:
@Bouncey The commit has been ammended - please review and tell me if further changes are needed.

Bouncey:
## LGTM :+1: 

Thanks for contributing :sparkles:

PR #96, opened 2017-07-27
Add Git introduction article

HKuz:
First draft of an introduction for the Git section. I included a "Sources" section because this article borrows info from the [Pro Git](https://github.com/progit/progit2) book (one that's used in Git's documentation), which has a Creative Commons with Attribution license.


QuincyLarson:
@HKuz This looks great. Thanks for writing this. And yes, I think your citation will be sufficient.

PR #97, opened 2017-07-27
Add SQL data type article

vanessasena:
I saw that there is also the topic SQL General Data Types, I don't know the difference with this one. I put everything here. Is it ok?


QuincyLarson:
@vanessasena Awesome! This is a great start. I've addressed the merge conflicts and am merging it. Thanks again for your contribution!

PR #98, opened 2017-07-27
full document entered

SteveChevalier:
Please review and let me know if changes or clarification is needed.
Thanks


QuincyLarson:
@SteveChevalier Nicely done! 

There was a minor merge conflict (@bouncey had just changed a few lines in every single stub on the master branch). I've resolved that merge conflict and also made a few minor fixes to your guide article:

- I made sure there was an empty line break between every element. This way images are always on their own line (before they were showing up in-line with the text).
- I turned the “ type quotes into " quotes.
- I don't think we need to point out that SQL stands for Structured Query Language in the title since most people just call it SQL. We can point that out on the top-level article on SQL.

Thank you again for investing the time and energy in planning and writing this first article. The hard part is over! The rest of these should be much easier now that you have a workflow.

I'm happy to help again if you run into any other questions. It was great talking with you earlier.

Cheers!

PR #99, opened 2017-07-27
Use Travis-CI to build site on PR

Bouncey:


Bouncey:
@BerkeleyTrue @QuincyLarson 

Travis runs out of memory when building the site also :disappointed:

BerkeleyTrue:
💩 

What are our other options?

Bouncey:
Heroku? AWS? 

Each build is currently taking around 10 minutes. My system has 8GB and can build it, each netlify container has 7GB, but I don't know how much of that is taken up with the docker image. 

This build process is only going to get longer and bigger.

QuincyLarson:
@Bouncey If you're OK with us using your system to build, then yes - let's use that. Otherwise we would need to pay DO or AWS $80/mo for a dedicated build server.

Bouncey:
I will see if I can run the build script on the same machine the search bar currently uses so it can be automated with a webhook. 

Other wise it will be a manual process on my machine at home

Bouncey:
I will only charge $79.95/mo 😆 😆 🤣 🤣

Bouncey:
Hmm, I must have been right on the limit yesterday when I last built the sight, with the new contributions pulled in I cannot build the sight any more. 

I don't think gatsby is right for us at the minute :disappointed:

Bouncey:
@BerkeleyTrue @QuincyLarson I think we need a call to hash this out. I will be available in two hours

Bouncey:
gatsbyjs/gatsby#1651

resolves this for now

PR #100, opened 2017-07-27
None relative markdown links get 'target' and 'rel' attrs

### LINKED ISSUE #30: Apply on all non-internal links: target="_blank" rel="nofollow"
Since internal links will use relative paths, we can just apply the target="_blank" and rel="nofollow" to any non-relative path URL. Is it possible to configure this within Gatsby?

Bouncey:
closes #30


QuincyLarson:
@Bouncey This looks solid. Thanks for running the script - the output is good proof that everything works.

PR #106, opened 2017-07-31
Updated the "A Href" attribute index.md with some information and an example

GuillermoGarrido:
Greetings, is this example good enough to use for the guide article?


QuincyLarson:
@GuillermoGarrido Yes - this is a great start. I'm going to go ahead and merge this because it's fine, and it's better than the stub we have now. 

If you can go into more detail with future articles, that would be even better. Thanks for your contribution!

PR #107, opened 2017-07-31
add info and example for articles

nelruk:
add iframe concept, href attribute and comment tag

### DIFF
--- a/src/pages/articles/html/a-href-attribute/index.md
+++ b/src/pages/articles/html/a-href-attribute/index.md
@@ -3,13 +3,8 @@ title: A Href Attribute
 ---
 ## A Href Attribute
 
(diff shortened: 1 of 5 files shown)


QuincyLarson:
@nelruk Excellent work! Thank you for contributing these articles! Merged!

texas2010:
oh dang. I noticed that link say https://www.freecodecamp.com.
more like it suppose to say https://www.freecodecamp.org.

nelruk:
about your https://github.com/freeCodeCamp/guides/pull/107#issuecomment-320791295 comment @texas2010 for me it works fine the link! :)

QuincyLarson:
@nelruk it still works (it gets redirected) but we should use freecodecamp.org going forward. No worries for now, though. We can do a global find-replace later.

texas2010:
yeah. not a big deal.  I know we could fix it later.

PR #110, opened 2017-08-03
Add article SQL insert into statement

vanessasena:


Bouncey:
Grate article @vanessasena :+1: 

Thank you :sparkles:

PR #114, opened 2017-08-06
Added jquery and bootstrap's javascript

nsuchy:
This pull request adds jquery and bootstrap's javascript. This change is needed for pages that show examples utilizing jquery and bootstrap. Without adding jquery and bootstrap's javascript showing examples in action is not a possibility.

### DIFF
--- a/src/html.js
+++ b/src/html.js
@@ -51,6 +51,28 @@ export default class HTML extends React.Component {
             }
             rel='stylesheet'
           />
+          <script
+            crossOrigin='anonymous'
+            integrity={
+              'sha384-xBuQ/xzmlsLoJpyjoggmTEz8OWUFM0/RC5BsqQBDX2v5cM' +
+              'vDHcMakNTNrHIW2I5f'
+            }
+            src={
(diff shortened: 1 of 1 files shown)


Bouncey:
@nsuchy thanks for the pull request.

I will need to investigate further to see if there will be any conflicts between these libraries and react/react-bootstrap.

We use `react-bootstrap` for our UI, and it accomplishes all aspects of `bootstrap.js` without `jQuery`. I understand the need for these libs to correctly show working examples.

@BerkeleyTrue Did you manage to integrate these into the challenge framework without conflict?

BerkeleyTrue:
@Bouncey on fcc, challenges are run in their own isolated iframe.  A challenge that requires jquery or bootstrap imports it into the iframe itself. This means it does not have a change to interfere with the rest of the site.

Bouncey:
I'll have to have a think about it.

nsuchy:
@Bouncey I'm not sure if it's out of scope or even a possibility but maybe we could add an "imports" feature to the markdown parser so if a certain library is needed it'll be imported just for that one page. Thoughts?

nsuchy:
Closed in favor of #274

PR #116, opened 2017-08-07
add information and example to the article

nelruk:
the example includes a .html document


QuincyLarson:
@nelruk Awesome - thanks for getting these articles off to a strong start with these explanations and examples.

PR #118, opened 2017-08-07
Added freeCodeCamp's favicon

nsuchy:
Just like the freeCodeCamp curriculum and forums - the guides website should have freeCodeCamp's favicon. This PR adds the favicon to provide a consistent user experience.

### DIFF
--- a/src/html.js
+++ b/src/html.js
@@ -5,6 +5,10 @@ import Helmet from 'react-helmet';
 import { TypographyStyle } from 'react-typography';
 import typography from './utils/typography';
 
+import fccMetaTags from './meta.js';
+
+const meta = fccMetaTags();
+
 export default class HTML extends React.Component {
   static propTypes = {
     body: PropTypes.string,
@@ -51,6 +55,7 @@ export default class HTML extends React.Component {
             }
             rel='stylesheet'
           />
+          {meta}
           <title>freeCodeCamp Guide</title>
           <TypographyStyle typography={typography} />
           {css}
--- a/src/meta.js
+++ b/src/meta.js
@@ -1,0 +1,233 @@
+import React from 'react';
+
+function fccMetaTags() {
+  return [
+    <link
+      href='https://s3.amazonaws.com/freecodecamp/favicons/favicon.ico'
+      rel='icon'
+      type='image/x-icon'
+    />,
+    <link
+      href='https://s3.amazonaws.com/freecodecamp/favicons/favicon.ico'
+      rel='favicon'
+    />,
+    <link
+      href='https://s3.amazonaws.com/freecodecamp/favicons/favicon.ico'
+      rel='shortcut icon'
+    />,
+    <link
+      href={
+        'https://s3.amazonaws.com/freecodecamp/favicons/' +
+        'favicon-16x16.png'
+      }
+      rel='favion'
+      sizes='16x16'
+    />,
+    <link
+      href={
+        'https://s3.amazonaws.com/freecodecamp/favicons/' +
+        'favicon-32x32.png'
+      }
+      rel='favicon'
+      sizes='32x32'
+    />,
+    <link
+      href={
+        'https://s3.amazonaws.com/freecodecamp/favicons/' +
+        'favicon-96x96.png'
+      }
+      rel='favicon'
+      sizes='96x96'
+    />,
+    <link
+      href={
+        'https://s3.amazonaws.com/freecodecamp/favicons/' +
+        'android-chrome-36x36.png'
+      }
+      rel='android-chrome'
+      sizes='36x36'
+    />,
+    <link
+      href={
+        'https://s3.amazonaws.com/freecodecamp/favicons/' +
+        'android-chrome-48x48.png'
+      }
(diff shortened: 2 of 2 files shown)


QuincyLarson:
@Bouncey what do you think about this? We want to add a ton of other meta data, as well, and this is important for SEO purposes. 

Here's what freeCodeCamp's current meta data partial looks like: https://github.com/freeCodeCamp/freeCodeCamp/blob/7805d74ea73881d1143474f3cb5a4e71c7bc1e06/server/views/partials/meta.jade

nsuchy:
@QuincyLarson Would you like similar metadata added to this project? I can implement if you'd like.

src/html.js
```diff
@@ -51,6 +51,17 @@ export default class HTML extends React.Component {
             }
             rel='stylesheet'
           />
+          <link
+            href={
```
Bouncey: We don't need the `{}` if it is just a string. He curly braces are there to help you do something directly in JavaScript, like string concatenation. We would concat a string if it took the line length over 80 characters.

src/html.js
```diff
@@ -51,6 +51,17 @@ export default class HTML extends React.Component {
...
+              'https://s3.amazonaws.com/freecodecamp/favicons/favicon.ico'
+            }
+            rel={
+            'icon'
```
Bouncey: As above

src/html.js
```diff
@@ -51,6 +51,17 @@ export default class HTML extends React.Component {
...
+          }
+            type={
+              'image/x-icon'
```
Bouncey: As above

Bouncey:
@nsuchy the majority of that meta partial is relation to the favicon and related themes. 

Taking the icon from freeCodeCamp/assets repo, you can use a favicon generator website to generate the HTML and the relevant files like a manifest.json and so on. 

Like in the partial above, can you make a `Meta` component which can be imported into `html.js`?

src/html.js
nsuchy: So confirm this isn't an always thing, you only use that method when it's longer than 80 characters. If that's the case I'll make the needed adjustment.

src/html.js
nsuchy: See above response.

src/html.js
nsuchy: See above response.

nsuchy:
@Bouncey to confirm you'd like me to adjust this to include all types of favicons?

Bouncey:
@nsuchy yes please

nsuchy:
@Bouncey Got it - I will be working on the changes and will post an update shortly.

nsuchy:
@Bouncey I've amended the requested changes into the commit - would you mind reviewing again?

Bouncey:
Oh wow! That makes the `html.js` file really noisy. 

Can you create a separate file with a function that returns and array of meta tags? I suggest an array because we cannot wrap them all in a `div` like a normal comonpent, no divs inside the head.

nsuchy:
I have setup a meta.js file, html.js imports it and runs the function at the proper location, and the meta tags for favicons are added to the page as required. This should clean things up, it's all been amended into the commit. Thoughts?

src/html.js
```diff
@@ -51,6 +53,7 @@ export default class HTML extends React.Component {
             }
             rel='stylesheet'
           />
```
nsuchy: Please see the changes I amended into the commit.

src/meta.js
```diff
@@ -0,0 +1,233 @@
...
+    />,
+    <meta
+      content='#FFFFFF'
+      name='msapplication-TileColor'
+    />,
+    <meta
+      content='/'
+      name='msapplication-TileImage'
+    />
+  ]);
```
Bouncey: Almost there, each `Component` in an array requires a `key` prop. This helps react know which component is which when it comes to rendering updates. All react components are Objects, so we can asign a unique key dynamically like in this example:

```js
const arrayOfComponetents = [
  <Component1 />,
  <Component2 />,
  <Component3 />
  //...
  ]
  .map((Component, i) => ({ ...Component, key: i });
```

Where `Component` is the react component we are working with currently and `i` is it's index in the array. We return a new Object (read `Component`) with a property of `key` set to the the `Component`'s index in the array.

This will give each `Component` a unique identifier that react can use to do the voodoo they do. :sparkles: :sparkles:

src/meta.js
```diff
@@ -0,0 +1,233 @@
+import React from 'react';
+
+function fccMetaTags() {
+  return ([
```
Bouncey: We don't need to wrap the array in `()`

nsuchy: Noted.

src/meta.js
nsuchy: Resolved.

nsuchy:
Please see the requested changes amended into the commit.

Bouncey:
## LGTM 👍 

I will pull this down to test tonight.

Bouncey:
Great work @nsuchy 

Thanks for contributing :sparkles: :sparkles:

PR #121, opened 2017-08-08
Create article for absolute value

codersc:
I hope this can at least be a rough outline for the absolute value article. The markdown I used seems to be wrong somehow, but I can't figure out why. For example, if I delete all the text from the article the preview still shows text in it. So, it doesn't seem to be giving me an accurate preview as far as I can tell. Also, I couldn't figure out how to remove the green vertical bars from the left side of the text that's showing in my preview, and I wasn't sure whether to add a section on the LaTeX syntax of the absolute value function. Maybe someone can edit the article, or just let me know how to fix it?

### DIFF
--- a/src/pages/articles/math/absolute-value/index.md
+++ b/src/pages/articles/math/absolute-value/index.md
@@ -3,11 +3,33 @@ title: Absolute Value
 ---
 ## Absolute Value
 
-This is a stub. <a href='https://github.com/freecodecamp/guides/tree/master/src/pages/articles/math/absolute-value/index.md' target='_blank' rel='nofollow'>Help our community expand it</a>.
+You can think of a number's absolute value as its distance from zero.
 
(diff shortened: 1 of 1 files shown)


Bouncey:
@codersc I think you are viewing the rich diff. The diff will show you what has been removed (highlighted red) and what has been added (highlighted green). The rich diff shows red/green bars in the margin.

[This is the actual preview of your article](https://github.com/codersc/guides/blob/7903c899140b288e2ffb63215be38c07dab2d1e2/src/pages/articles/math/absolute-value/index.md)

QuincyLarson:
@codersc Hi Kevin, this is an excellent article. Thank you for your contribution! And congratulations on your first contribution to this repository! We would welcome more like this :)

codersc:
@QuincyLarson Hey, thanks a bunch!

PR #132, opened 2017-08-09
Update Add Fractions with Unlike Denominators article

codersc:


QuincyLarson:
@codersc Thanks for your contribution. Any additional detail you can add later will be helpful, too. We are working on getting MathJax working, and at which point we can render these equations as notation rather than images.

codersc:
Alright @QuincyLarson, sounds good. MathJax seems like a great option for these articles.

QuincyLarson:
@codersc We will try to get MathJax working on guides separately. We should for all intents and purposes treat guides as separate from the freeCodeCamp platform. 

So I would recommend using images of equations instead, which we can later replace with MathJax.

PR #146, opened 2017-08-10
Adjust Travis-CI to emulate the netlify build process

Bouncey:


Bouncey:
Travis times out, the HTML build alone is ~20 minutes

PR #158, opened 2017-08-15
Improve Redux implementation performance

### LINKED ISSUE #156: Performance issues while running locally.
Sometimes while running the repo locally, it gets stuck at the point where it reads `Bootstrap finished`.

Bouncey:
closes #156 

Previously, only SideNav was connected to redux, this meant there was a long chain of props and checking against the store, now each Panel has access to the store.

gatsby has released a breaking change in one of its' packages, I have locked down the `package.json` to explicit versions re: gatsby for the time being.


QuincyLarson:
Awesome - this is a massive improvement in performance.

PR #170, opened 2017-08-19
Bring in Search with type ahead

Bouncey:


Bouncey:
@QuincyLarson this will need a good QA

src/head/styleSheets.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+
+const stylesSheets = [
```
systimotic: I think this should be `styleSheets` instead of `stylesSheets`

Bouncey: Good catch

Bouncey:
Thanks for the great review @systimotic I have taken all of your points on board.

There is no distinction for react files, as long as you make webpack aware that some will be be `js` and some will be `jsx` which is why you would see `test: /jsx?/` in a `webpack.config`.
`html.js` is a `js` because it is a `gatsby` thing. Changing is to `jsx` would mean submitting PR to gatsby possibly.
Personally, I like the `jsx` extension because I get a react icon next to it in my editor.

I have a quick chat with @QuincyLarson last night am I am going to try out a "Search Ahead" option. Meaning I take away the type ahead's and the search button, so all you have is an input. As soon as a user types something, we display search results on a dedicated search page.

Bouncey:
closed in favour of #181

PR #171, opened 2017-08-19
Change dev npm script to run on Windows

### LINKED ISSUE #169: Running on Windows
The `dev` npm script does not work on Windows. It directly refers to `./node_modules/.bin/gatsby`, which is a bash file. It does not run on Windows. If I change it to the `.cmd` version, it tries to run it as if it's a Javascript file, so it fails too.

I noticed that the all that the bash file does. I'm wondering whether the extra available memory transfers to that process. @Bouncey Have you seen that this was working?

I have a couple of ideas on how we could potentially fix this.
1. Instead of referencing the bin file, we could reference the js file that the bin file references directly.
2. We could have one script that runs gatsby normally, then another script that runs that script with more memory available. I'm not sure if this would work.
```json
"dev-low-memory": "gatsby develop",
"dev": "npm run dev-low-memory -- --max_old_space_size=7168"
```
3. We could try creating a node script that runs Gatsby as a child process, and create an npm script that runs that node script with more available memory. Again, I'm not sure if this would work.

systimotic:
Addresses #169

In a DM on Gitter, @Bouncey mentioned the extra memory was not required for the dev script. I have removed it so that the dev script now also works on Windows.

I think this closes #169

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,7 +59,7 @@
   "scripts": {
     "_comment": "ONLY RUN BUILD IF YOUR SYSTEM HAS 8GB OR MORE OF RAM",
     "build": "node seed && node --max_old_space_size=7168 ./node_modules/.bin/gatsby build",
-    "dev": "node seed && node --max_old_space_size=7168 ./node_modules/.bin/gatsby develop",
+    "dev": "node seed && gatsby develop",
     "lint": "eslint --ext .js,.jsx --ignore-path .gitignore .",
     "normalise": "node _normaliseArticles.js",
     "test": "echo \"Error: no test specified\" && exit 1"


package.json
```diff
@@ -59,7 +59,7 @@
   "scripts": {
     "_comment": "ONLY RUN BUILD IF YOUR SYSTEM HAS 8GB OR MORE OF RAM",
     "build": "node seed && node --max_old_space_size=7168 ./node_modules/.bin/gatsby build",
-    "dev": "node seed && node --max_old_space_size=7168 ./node_modules/.bin/gatsby develop",
+    "dev": "gatsby develop",
```
Bouncey: We would still need to run `node seed` before running gatsby. The seed script builds a JSON which is used to create the side navigation.

systimotic: @Bouncey Whoops! Wasn't paying enough attention when I made this change. Has now been fixed

Bouncey:
I can't test this because the only windows machine I have is from work, and they don't like me installing stuff. 

@systimotic can you just confirm that this works on windows?

systimotic:
@Bouncey I can confirm that this works on my Windows 10 laptop, both in the default command prompt and in Git Bash. Only issue is that I can't seem to stop the process in Git Bash, but that's not related to this.

Bouncey:
Great work @systimotic 👍 

Thanks ✨

PR #192, opened 2017-08-22
Replace typography package with custom styles

systimotic:
This PR replaces the typography npm package with custom CSS styles.

The typography plugin was adding margin to elements it shouldn't be touching, and wasn't adding a whole lot.

Feel free to squash these commits on merge, I kept the separate for the PR to help explain which changes are for which reason.

Something I noticed after making these changes is that we really need some guidelines on the size of titles. Some articles use h1 as their first heading, most use h2, some use h2 for all headings. It's a bit of a mess.


Bouncey:
Looking good to me, I'll pull this down to test today

Bouncey:
## LGMT 

Thanks @systimotic :sparkles:

PR #194, opened 2017-08-23
Tweak styles

systimotic:
This PR tweaks some styles, removes some unnecessary stuff and move some files around.

~It's currently marked as [WIP] because it depends on #192. 2bf1197 is the only commit from this PR.~

I think this is another step in the right direction. I hope to work on the styling even more in the future.


src/css/sideNav.module.css
```diff
@@ -1,12 +1,4 @@
-@keyframes grow {
```
Bouncey: This file need to be renamed to `SideNav.module.css`

Bouncey:
I fixed a merge conflict for you also

src/css/sideNav.module.css
systimotic: @Bouncey Whoops! Turns out I had an issue I've had before: git isn't case sensitive in its rename detection. This worked locally because I had actually renamed the file, but git didn't count it as a change (and I didn't notice). My previous workaround was to delete, commit, then add back in. I just did a search to see if there was a better way. Turns out you can use `git mv oldname newname` to do this.

systimotic:
@Bouncey I updated the PR. I did a rebase to remove the extra commit and the merge conflicts. I commented on your review explaining of why the file wasn't renamed previously 😅

Bouncey:
@systimotic can you rebase this and see if there is anything I missed?

systimotic:
Took me a while! Looks like everything has been merged correctly 👍

PR #196, opened 2017-08-24
Remove TypeAhead

systimotic:
The TypeAhead component was still present in the code, but it did not appear to be used. This PR removes that component and cleans up the syntax in a few places.

I can not say with 100% certainty that this doesn't break anything, as freecodecamp.duckdns.org is unreachable.


Bouncey:
## LGTM

Thanks @systimotic :sparkles:

PR #198, opened 2017-08-25
Update index.md for Adding and Subtracting Square Roots Article

codersc:


HKuz:
LGTM! 🤖 🚀 ☄️  Thanks @codersc! I added some temporary alt text to the images for accessibility - sounds like the long-term plan is to ultimately replace them entirely with LaTeX formulas when MathJax is integrated 😄

PR #201, opened 2017-08-25
Update Chickens versus Pigs

kailanjian:
Article was a stub, added content and removed stub mark.


HKuz:
LGTM! Thanks for your patience, @kailanjian! 🐷 🐔 🥓 🥚

PR #205, opened 2017-08-27
Add pre-commit message

systimotic:
The first time I created a commit I though it wasn't working. I didn't know that there was a pre-commit script, and nothing was happening, so I assumed something was wrong. This PR adds a message warning of the pre-commit script and that it can take a while to complete.

I tested this in Git Bash and the default command prompt on Windows 10.


Bouncey:
## LGTM 

Thanks @systimotic :sparkles:

PR #208, opened 2017-08-28
Update 404 page

systimotic:
Updated the styling and content of the 404 page to something that better fits the site.

I'm sorry for touching the styles 😓


Bouncey:
I agree with this PR, but we need to sort out our production CSS, then this will likely need a rebase

systimotic:
@Bouncey I updated this, should be ready now 🙌

src/css/main.css
```diff
@@ -56,7 +56,7 @@
   margin-bottom: 50px;
 }
 
-/* NoResults */
+/* ErrorPage */
```
Bouncey: ErrorPage doesn't refer to anything. Maybe it could be `NoResults/FourOfFour`

src/pages/404.jsx
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import Button from 'react-bootstrap/lib/Button';
+import Link from 'gatsby-link';
+
+function fourOhFour() {
```
Bouncey: This was my original mistake. React component declarations need to start with an uppercase char, else it will be parsed as HTML and not work as expected

So this should be `FourOhFour`

src/pages/search/components/NoResults.jsx
```diff
@@ -12,7 +12,7 @@ function NoResults(props) {
     <div className='flexWrapper'>
       <div className='verticalAlign'>
         <h3>{ helperHeader }</h3>
-        <p>Trying searching for something, or</p>
+        <p>Try searching for something else, or</p>
```
Bouncey: Please remove the `else` or add some logic so it reads well with the options for `helperHeader`

src/pages/404.jsx
```diff
@@ -0,0 +1,27 @@
...
+            >
+            Go to the homepage
+          </Button>
+        </Link>
+      </div>
+    </div>
+  );
+}
+
+fourOhFour.displayName = 'fourOhFour';
```
Bouncey: `'FourOhFour'`

src/pages/404.jsx
```diff
@@ -0,0 +1,27 @@
...
+
+export default fourOhFour;
```
Bouncey: `FourOhFour`

src/pages/404.jsx
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
```
Bouncey: Does gatsby pick this up if it is a jsx file?

systimotic: Seems to work locally, but haven't tested in a build. According to the docs, it should work. https://www.gatsbyjs.org/docs/add-404-page/ 
If you prefer to stick with .js or switch the others to jsx as well, let me know

src/pages/search/components/NoResults.jsx
systimotic: This one is tricky. I did think about this a little bit.

In the case that there is no search input, I thought that "Try searching for something" could come across as a sarcastic remark (don't know what that says about me 😅). 

I'm not sure what would be ideal here. How about "Please try searching for something using the search bar above, or" for the empty input cases?

Bouncey: How about `You can try searching for something, or`?

systimotic: Sorry, I was being too perfectionistic 😅

systimotic:
@Bouncey I just updated this again 🙂

Bouncey:
## LGTM :+1: 

Thanks @systimotic :sparkles: :sparkles:

PR #209, opened 2017-08-28
Add syntax highlighting

systimotic:
The `gatsby-remark-prismjs` package was already in the `package.json`, but it wasn't configured. I think syntax highlighting is nice to have for a site like this, so I decided to set it up. 

I also went through some of the packages to see if they were being used and removed the ones that weren't.

I tested this with the SQL Average Function article, and it worked great.

PrismJS does require you to specify the language. A JS code block would be written like the following

```markdown
```javascript
const prism = 'a popular syntax highlighting library';
```(placeholder)
```
without the `(placeholder)`, that's there because it's the only way GitHub will let me escape the backticks.

### DIFF
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -5,7 +5,6 @@ module.exports = {
   },
   plugins: [
     'gatsby-plugin-react-helmet',
-    'gatsby-transformer-remark',
     {
       resolve: 'gatsby-source-filesystem',
       options: {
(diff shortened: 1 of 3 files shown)


Bouncey:
Does it work with ` ```js`

systimotic:
@Bouncey I just tested this, it does!

Bouncey:
## LGMT :+1: 

Thanks @systimotic :sparkles:

PR #210, opened 2017-08-29
Add an index page

### LINKED ISSUE #148: Create a landing page
Create a good index.md file for the root directory (that shows when you navigate to Guides directly.

This file should be situated in  `src/pages/ `.

systimotic:
I decided to write an index page so I could have something nicer to look at while developing. I'm not the greatest writer, so improvements would be more than welcome.

My personal problems with this page:
- The list is kind of nice, but also a bit redundant with the navigation there. It's not as good as the nav, because it doesn't show allll the things. I did only pick things that already have at least a couple of articles.
- The whole thing feels a bit too marketing-y
- I'm not sure if I can call this its own site as its URL will be freecodecamp.org/guides

Kinda maybe sorta closes #148

### DIFF
--- a/src/pages/articles/index.md
+++ b/src/pages/articles/index.md
@@ -1,6 +1,26 @@
 ---
 title: Articles
 ---
-**AlphabetJS** is the latest and greatest Javascript Framework. It's
-amazing and wonderful and you'll probably love it more than all of the
-other ones.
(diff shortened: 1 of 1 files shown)


QuincyLarson:
@systimotic This is a great start. I'll merge it and I'm sure it will evolve over time.

PR #211, opened 2017-08-29
Refactor CSS Modules :cry: 

Bouncey:


systimotic:
An excerpt from a DM on Gitter earlier:
Why are we switching in the first place? If the issue is only the original problem, here's my reasoning on why switching doesn't make sense: If we were to update, we could use CSS modules again. However, we can't seem to update because it gives new, unrelated issues. Sticking with the current fork is not an issue because if we update to the latest version, the fix from the fork will still be there. Ideally, we stick with the fork and open an issue on Gatsby's GitHub for the updating problems.

I'll try to find time to review this PR tomorrow. Would love to hear your thoughts on the message above.

PR #215, opened 2017-08-31
Move babel presets to dependencies

Bouncey:
Whilst building in netlify, it looks like dev-dependencies are not installed. This causes an error and the build process exits with a non-zero exit code, due to `es2015` not being found.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,6 @@
   "author": "freeCodeCamp",
   "dependencies": {
     "axios": "^0.16.2",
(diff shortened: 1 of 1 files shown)


Bouncey:
No, it can remain a devDependency

systimotic:
👍 If you move `babel-cli` back to `devDependencies`, I think this is good to merge.

Bouncey:
My bad, I thought I had left it there.

PR #216, opened 2017-08-31
Have travis run build on PR

Bouncey:
This step will help our deployment process. 

At times errors are only found on `build`. 

We should also use `yarn` in travis to keep things constant across platforms


Bouncey:
Nevermind. It takes too long for Travis

PR #221, opened 2017-08-31
Format SQL reserved words correctly

### LINKED ISSUE #143: The function titleify should account for SQL key words
`function titleify() {}` found [in this file](https://github.com/freeCodeCamp/guides/blob/master/seed/utils/titleify.js) should account for SQL keywords like 'COUNT', `AND`, 'AVG', 'INSERT', `WHERE` and so on.

I envisage `titleify` checking if the title contains the word `sql`, where is will then call another function to deal with SQL keywords.

Perhaps @SteveChevalier or @vanessasena could provide a list of keywords as a reference?

Bouncey:
closes #143


Bouncey:
How about this for an idea.

We reduce normalise down to a script for formatting links in articles (target='_blank'... ), stubbing new articles and things like that. All edits after that become the concern of the community.

This will greatly reduce our workload in having to apart `titleify` for each use case.

Bouncey:
closed in favour of #242 as it implements the above [comment](https://github.com/freeCodeCamp/guides/pull/221#issuecomment-326827674)

PR #222, opened 2017-08-31
Rename top level Math to Mathematics

### LINKED ISSUE #159: Incorrect article sorting for Side Nav
Articles that are meant for `articles/math` are also appearing in `articles/javascript/standard-objects/math` in the Side Nav

I suspect that `createNavData` requires a refactor.

Bouncey:
closes #159 

The toplevel `Math` category was being incorrectly sorted i to `javascript/standard-objects/math`

Renaming the top level dir solves this for now. Gatsby is making inroads to having all pages available to the Layout component, which will mean we wont have to do this sort of sorting any more. At that time, we can rename this back to `Math` if required


systimotic:
@Bouncey I started working on a refactor of `_createNavData.js` today. I won't be able to finish it today, but I got to a point where I feel like I'd probably be able to get it working. My solution involves generating `navData` with a structure like the following:

```js
{
  ...
  "javascript": {
    ...
    "fullChildren": {
      "standard-objects": {
        ...
        "fullChildren": {
          "math": ...
        }
      }
    }
  },
  "math": {
    ...
  }
  ...
}
```
I kind of like it, but it's definitely not ideal. If we don't need this at all in the future, is it worth it to try to make this fix work 100%, or should we go with your workaround?

Bouncey:
I'd be interested to see what you have, if it's not too much hassle?

systimotic:
Sorry for the delay. The refactor probably isn't worth it if Gatsby is going to offer this by default soon. My refactor isn't done yet, and we're close to launch, so I'll review this tomorrow and merge it if it works. This will cause conflicts with https://github.com/freeCodeCamp/guides/pull/242/ I think it's easiest to merge that first, then reapply the changes from this PR again.

PR #223, opened 2017-08-31
Update URL in README

Bouncey:


Bouncey:
HTTPS yes, it will happen, just need to get the search on HTTPS first so it will still work

guide vs guides? I think the repo should have been guide also, but someone got excited when creating it @QuincyLarson

systimotic:
🔒 👍 
Singular suggests to me it is a guide to freeCodeCamp, plural suggests they are guides from freeCodeCamp, but 🤷‍♂️ not a big deal

PR #227, opened 2017-09-01
Search over HTTPS

Bouncey:
Because https

This will also enable us to serve the guide over HTTPS when the time comes


Bouncey:
I have to adjust the nginx config for this. I'll ping you when I have it sorted.

Bouncey:
This should be working again

PR #93, opened 2017-07-25
Navbar height fixed to 40px

invinciblycool:
closes #68

PR #229, opened 2017-09-01
Change navbar height to 40px

### LINKED ISSUE #68: Fix navbar height
The navbar should be 40px tall, which is what we're planning to use throughout freecodecamp.org. The logo and search box should be vertically centered within the nav bar.

### LINKED ISSUE #161: Navbar layout not displaying properly in IE11
guide layout on IE11

systimotic:
Closes #68 
Closes #93 
Closes #161

Regarding #161: here's a screenshot from this PR in IE11
iexplore_2017-09-01_10-26-24
As you can see, it's not centered, but it's a big improvement over the current situation.

The react-bootstrap `Navbar` component automatically adds a `container` for the children, which I then can't easily style. I chose to make this more explicit in the markup rather than adding global CSS for `.navbar > .container`. If the CSS solution is preferred, I can switch to that.

In Firefox, the logo isn't perfectly centered. This is because the image we're using has some empty space on the bottom. It looks like Chrome tries to do something to mitigate this, but it makes the image a bit blurry.

Responsiveness isn't great. I hacked something together to make it better than before, but it's still bad. This is probably something we should tackle before launch, right? I'm assuming we'll probably need a button for the menu in the navbar for screens smaller than 992px, where the SideNav and content don't fit next to eachother anymore, so it probably doesn't make sense to try to make it work properly in this PR. We could probably take some more inspiration from devdocs.io here.


src/layouts/Layout.jsx
```diff
@@ -30,31 +30,29 @@ class Layout extends React.PureComponent {
...
-              </Col>
-              <Col md={ 9 } xs={ 12 }>
-                <SearchBar />
-              </Col>
-            </Navbar>
-          </Row>
-        </Grid>
+        <nav className='navbar navBar'>
+          <Grid className='navContainer'>
+            <Col className='hidden-xs' sm={ 3 }>
```
Bouncey: I'm not sure about not showing the logo on smaller devices

systimotic: Absolutely agree. This is definitely not ideal, and this is what I meant about the responsiveness still not being great.

The reason I went with this is that there isn't enough space to show both the logo and the search bar. devdocs.io solved this by not showing a logo. freeCodeCamp.org solved this by only showing the logo and a menu button.

Ideally, we do something similar to the devdocs solution. Have the searchbar right there, have a button for the nav. There still wouldn't be space for a logo. My idea was to go with this for now to make it usable, then make it look better in a PR that fixes the navigation on smaller screens.

Bouncey: @QuincyLarson hiding the logo on mobile?

Bouncey: From @QuincyLarson 
> For the guide, we want the logo to be visible because one of our goals is to raise awareness of our community to people arriving from Google searches.

systimotic: @Bouncey I absolutely agree, but I don't know how it would be possible to show both the search bar and the logo on a small screen. How about putting the glyph version in place of the full logo? https://freecodecamp.github.io/design-style-guide/#logo
It's not the perfect branding because it doesn't show the name, but it might be the best compromise.

QuincyLarson: @Bouncey perhaps on smaller breakpoints we would need to show the glyph. I imagine even an 80px wide search bar would suffice though. We have no other navigation.

systimotic:
@Bouncey Sorry for the wait. I've updated this PR to show the glyph on smaller screens. 

I removed the bootstrap code from the navbar. Because the glyph does not fit into the column system, I would have to either create HTML for a whole separate navbar that would show on smaller screens, or I would have to override all of the Bootstrap styles in a breakpoint.
While my solution means I had to redo some Bootstrap code, I think it's the most efficient one.

I could keep the big logo for a bit longer, but it's hard to judge where the switch should be made. I made it at the point where the navbar would have to be smaller than 75%, which is also the point where the navigation breaks, so it's the point where a menu button would be displayed, meaning that in the future the search bar will be smaller at this width.

Bouncey:
The code looks good, but it's going to conflict with #252, or the other way round. I'll have to pull this down to look at it

systimotic:
Yeah, it will definitely conflict, but I think it shouldn't be too bad. I think it would be the easiest to resolve if me merge the service worker first, then work this into that.

QuincyLarson:
Hi @systimotic Have you had a chance to take a look at this and address the conflicts?

MrAsrani1312:
The build shows now conflict now. Wonder if I could merge 🤔

MrAsrani1312:
Ok....got an update with the conflicts...sorry for the above post

QuincyLarson:
@Bouncey Can you take a look at this when you get a chance? You're the most familiar with this JSX - perhaps you could QA and resolve merge conflicts if this looks good?

Bouncey:
Thanks @systimotic, sorry for the long delay. These changes look great!

QuincyLarson:
Thanks, @systimotic!

PR #237, opened 2017-09-03
Inform users of no support for their browser

### LINKED ISSUE #217: Promise is not supported by IE 11
@QuincyLarson From which browser version are we supporting?

Currently the search will not work in IE11. 

We could:

- Add a polyfill for Promise's
- Refactor to not use Promise's
- Inform the user that the search page does not support their browser after doing a `'Promise' in window` check, and suggest they try chrome/firefox/edge

Personally I am not a fan of shipping polyfill to everyone regardless, and I like Promise's. So I would vote for the 'No support search page' and suggestion of upgrading.

Bouncey:
screenshot from 2017-09-03 23-31-49
screenshot from 2017-09-03 23-33-36
closes #217

### DIFF
--- a/src/pages/search/components/NoSupport.jsx
+++ b/src/pages/search/components/NoSupport.jsx
@@ -1,0 +1,67 @@
+import React from 'react';
+import FontAwesome from 'react-fontawesome';
+import { Row, Col } from 'react-bootstrap';
+
+function NoSupport() {
+  return (
+    <div className='flexWrapper'>
+      <div className='verticalAlign'>
+        <h3 className='colourDarkGrey text-center'>
+          This page is not supported in your browser
+        </h3>
+        <p className='colourDarkGrey'>
+          {
+            'We suggest upgrading to the latest version of one of the ' +
+            'browsers below'
+          }
+        </p>
+        <hr />
+        <Row className='row-fluid'>
+          <div className='flexSpaceAround'>
+            <Col className='flexSpaceAround' sm={ 4 }>
+              <a
+                href='https://www.microsoft.com/en-gb/windows/microsoft-edge'
+                rel='nofollow'
+                target='_blank'
+                >
+                <FontAwesome
+                  className='browserIcon'
+                  name='edge'
+                  size='4x'
+                />
+              </a>
+            </Col>
+            <Col className='flexSpaceAround' sm={ 4 }>
+              <a
+                href='https://www.google.com/chrome/browser/desktop/index.html'
+                rel='nofollow'
+                target='_blank'
+                >
+                <FontAwesome
+                  className='browserIcon'
+                  name='chrome'
+                  size='4x'
+                />
+              </a>
+            </Col>
+            <Col className='flexSpaceAround' sm={ 4 }>
+              <a
+                href='https://www.mozilla.org/en-GB/firefox/new/'
+                rel='nofollow'
+                target='_blank'
+                >
+                <FontAwesome
+                  className='browserIcon'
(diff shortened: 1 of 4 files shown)


src/pages/search/components/NoSupport.jsx
```diff
@@ -0,0 +1,67 @@
+import React from 'react';
+import FontAwesome from 'react-fontawesome';
+import { Row, Col } from 'react-bootstrap';
+
+function NoSupport() {
+  return (
+    <div className='flexWrapper'>
```
systimotic: This uses `.flexWrapper` and `.verticalAlign` from `NoResults/404`. Should we move those classes to global or some section specific to these kind of error pages?

src/pages/search/components/NoSupport.jsx
```diff
@@ -0,0 +1,67 @@
...
+      <div className='verticalAlign'>
+        <h3 className='colourDarkGrey text-center'>
+          This page is not supported in your browser
+        </h3>
+        <p className='colourDarkGrey'>
+          {
+            'Either upgrade to the latest version or download one of the ' +
```
systimotic: IE11 is the latest version of IE, but it doesn't support promises. How about "We suggest upgrading to the latest version of one of the browsers below"?

src/pages/search/components/Results.jsx
```diff
@@ -34,27 +40,6 @@ function mapDispatchToProps(dispatch) {
   }, dispatch);
 }
 
-const faNames = {
-  challenge: 'free-code-camp',
-  guides: 'file-text',
-  youtube: 'youtube-play'
-};
-
-function FAIcon({ name }) {
```
systimotic: Good call on refactoring this 👍

src/pages/search/components/NoSupport.jsx
Bouncey: I thought edge was the latest version of IE? I linked to the edge download page, so if someone was on an older version of IE they would likely click the edge logo.

src/pages/search/components/Results.jsx
Bouncey: I don't know why I put a wrapper around it in the first place 😕

src/pages/search/components/NoSupport.jsx
Bouncey: Probably, but I think it is out of scope for this PR. I noticed the latest version of Gatsby comes with a CSS Modules plugin, so I would like to give that a go if I can get the latest version of Gatsby to work here

Bouncey:
What is the uncaught error? I cannot dev on a machine with IE on it. So I was half blind building this

src/pages/search/components/NoSupport.jsx
Bouncey: And I think it will be more like a CSS utils file, so `.flexSpaceAround` could go in it too

src/pages/search/components/NoSupport.jsx
systimotic: According to Microsoft, Edge and IE are different browsers. The logo's are similar enough that this might work, though. I'm still in favor of changing the copy.

Bouncey:
@systimotic copy updated

systimotic:
iexplore_2017-09-07_15-00-00
You can get the same error with more deletes by running `delete Promise` in the console in Chrome. It doesn't seem to impact functionality though. Hasn't been created by this PR, and doesn't functionality, so I think it's OK to leave it.

PR #242, opened 2017-09-05
Remove the articles directory

### LINKED ISSUE #143: The function titleify should account for SQL key words

### LINKED ISSUE #147: Remove the articles directory
Currently, Gatsby sorts the guide pages by their position in the `src/pages` directory. The way we currently have the directory structured, all of your paths are prefixed with `articles`, because we have all of our pages in an `articles` directory.

When release in the wild, this guide will be found at `freecodecamp.org/guide` and the way things are, navigation around our site would be `freecodecamp.org/guide/articles/<some-category>/<some-article>`.

We would like to have URLs like `freecodecamp.org/guide/<some-category>/<some-article>`, removing the `articles` path level.

For this to be accomplished, all articles need to be removed from `articles` and placed into `src/pages/`. `_normaliseAriticles` and `createNavData` will need to be refactored to remove the dependency on the `articles` dir.

Bouncey:
I have taken `_normaliseArticles` in the direction of 'it will stub new articles and format external links' as discussed in #221 

closes #143 
closes #147


Bouncey:
When I had the index file as a `.md` Gatsby just would not take it, it said there was not a page at `/`. So I converted it to a component. 

I have made the decision that `_normaliseArticles` will only stub new articles and format external links of existing articles. I think the formatting of the `title` prop should be left to the community. Every time there is a new paradigm added to the guide, titleify will need to be refactored, potentially. 

This can be merged without breaking the production site, I have disabled CD for the time being, so for this case I can synchronise the `search` update with the new guide going live manually

systimotic:
@Bouncey Hmm, but the `title` prop is unused, right? The title from `navData` is what's displayed in the navigation, and that's generated by taking the filename and running it through `titleify`. The title displayed on the page was taken from the content and wasn't being altered by the normalising script.

Also, merge conflicts 😢

Bouncey:
The title prop is used by `react-helmet` to update `document.title`,

I completely forgot about the SideNav using this also

Bouncey:
@systimotic `createNavData` no longer uses `titleify`, It takes the title directly from the frontmatter

systimotic:
So I was stupid and merged #222 before this. It doesn't show any merge conflicts, but merging would mean the articles directory would be back again, containing the `mathematics` folder, and the `math` folder would also be back. I have attempted to fix this. Let's hope this goes well. If not, not to worry, I have a backup branch, so I can restore it if necessary.

Bouncey:
Let me do a rebase first

systimotic:
Sorry @Bouncey! Didn't see your message before I pushed. A rebase did fix the problem, although it did give some merge conflicts. The changed articles then also needed to have the stubs updated too. Can you check my changes?

Bouncey:
## LGTM :+1:

invinciblycool:
Awesome!! :+1: :)

PR #247, opened 2017-09-06
Begin TypeScript guides

mtrivera:
This is the start of TypeScript guides for freeCodeCamp guides.

### DIFF
--- a/src/pages/articles/typescript/array-type/index.md
+++ b/src/pages/articles/typescript/array-type/index.md
@@ -1,0 +1,20 @@
+---
+title: Array Type
+---
+
+# Array Type
+
+You have full access to arrays in TypeScript.
+Arrays can be written two different ways in TypeScript:
+
+__DataType[]__  
+DataType followed by square brackets `[]`
+```typescript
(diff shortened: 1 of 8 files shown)


systimotic:
Thanks for contributing! I don't have time for a full review right now, so I'll leave that up to someone else. I do want to share a concern: I'm not sure how this would work on a Windows system, as the `?` is an illegal character in the file system.

Bouncey:
@mtrivera the articles look great, thanks for taking the time. However I second @systimotic's concern. But for a different reason. The directory structure in `src/pages/` will be the URL slug for the article it holds, which means we cannot have `?` in the folder title as that would be a URL query and could very well break things. 

You can certainly put the question mark the the `title` field though. 👍

mtrivera:
Thanks for the feedback. None of this is mentioned in the **Article Style Guide**. Maybe it should discussed be in a separate **Directory Style Guide** or appended to the **Article Style Guide**.

systimotic:
Good point @mtrivera! I've created an issue that addresses this: #250. Would you be interested in fixing this?

mtrivera:
@systimotic Sure

mtrivera:
See #254

Issue #250, opened 2017-09-07
Mention title's use in URL in style guide


systimotic:
The URL of an article is created from the folder title. Because of this, the folder title shouldn't contain any special characters, apart from dashes.
This should be mentioned in the style guide in the `README.md`

invinciblycool:
Opened #255. Let me know if any corrections are needed.

Bouncey:
closed via #253

PR #253, opened 2017-09-07
Add folder style guide & fix guide example link

mtrivera:
-  Added folder style guide to README.md #250. 
-  Fixed the guide article example link to point to 'pages'
- Changed 'articles' to 'pages' folder after contributing GIF

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,6 @@ This repo is where we plan and maintain these guide articles, which we then host
 - [What are Guide articles?](#what-are-guide-articles)
 - [What can I write an article about?](#what-can-i-write-an-article-about)
 - [How to contribute](#how-to-contribute)
-- [Folder style guide](#folder-style-guide)
 - [Article style guide](#article-style-guide)
 
 ## What are Guide articles?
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -63,6 +64,30 @@ yarn run dev
 
 In this project we are using `yarn` because `netlify` builds our site with `yarn`.
 
+### Folder style guide
+
+W> For new tech topics, you will need to create a new folder.
```
systimotic: I'm curious where the `W>` and `T>` are from. This uses GitHub Flavored Markdown, which doesn't support that. More information on the syntax: https://guides.github.com/features/mastering-markdown/#GitHub-flavored-markdown

README.md
```diff
@@ -63,6 +64,30 @@ yarn run dev
 
 In this project we are using `yarn` because `netlify` builds our site with `yarn`.
 
+### Folder style guide
```
systimotic: Instead of an entire "Folder style guide" section, I think this would be better under the "Article title" section.

README.md
```diff
@@ -63,6 +64,30 @@ yarn run dev
...
+W> \ (backslash)
+W> | (vertical bar or pipe)
+W> ? (question mark)
+W> * (asterisk)
+W>   (whitespace)
+
+>[What characters are forbidden in Windows and Linux folder names?](https://stackoverflow.com/questions/1976007/what-characters-are-forbidden-in-windows-and-linux-directory-names#1976050)
+><cite>Christopher Oezbek</cite>
+
+T> The dash character `-` is the ONLY accepted character in folder names other than letters and numbers.
```
systimotic: I think we can shorten this entire section to something like this:
"The folder name is used in the URL, so only use dashes, numbers, and lowercase letters for it."

README.md
mtrivera: https://webpack.js.org/writers-guide/ 
I thought it was universal but I guess not

Bouncey: That is some nice formatting! What flavour of markdown would that be?

mtrivera:
Made changes and added folder name examples

Bouncey:
Great work! 👍 

Thanks @mtrivera 
✨  ✨

systimotic:
Awesome, thanks @mtrivera!

PR #261, opened 2017-09-13
Added python article to convert string to integers

invinciblycool:
Let me know if any corrections or formatting edits are needed.
Will add more articles as soon I get tips on this article.

### DIFF
--- a/src/pages/python/how-to-convert-strings-into-integers-in-python/index.md
+++ b/src/pages/python/how-to-convert-strings-into-integers-in-python/index.md
@@ -46,8 +46,8 @@ Output:
 
 But you should keep in mind some special cases:
 
-1. A floating point(an integer with fractional part) as an argument will return the smallest nearest integer.
-   For example : `print(7.9)` will print `7`.
+1. A floating point(an integer with fractional part) as an argument will return the float rounded down to the nearest whole integer.
+   For example : `print(int(7.9))` will print `7`.
    Also `print("7.9")` will result an error since the string is an invalid argument to convert to an integer.
 
    ```py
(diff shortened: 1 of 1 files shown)


src/pages/python/how-to-convert-strings-into-integers-in-python/index.md
```diff
@@ -3,13 +3,69 @@ title: How to Convert Strings into Integers in Python
...
+print(age_int+2)
+```
+Output:
+```py
+20
+```
+
+But you should keep in mind some special cases:
+
+1. A floating point(an integer with fractional part) as an argument will return the smallest nearest integer.
```
Bouncey: From the doc you linked at the end of the article

> For floating point numbers, this truncates towards zero.

So my understanding is `int(7.9) == 7.0 == 7`

> return the smallest nearest integer

Maybe `return the float rounded down to the nearest whole integer` or something along those lines would be clearer

src/pages/python/how-to-convert-strings-into-integers-in-python/index.md
```diff
@@ -3,13 +3,69 @@ title: How to Convert Strings into Integers in Python
...
+   For example : `print(7.9)` will print `7`.
```
Bouncey: Am I getting confused over `print()` and `int()`?

Wouldn't `print(7.9)` just print 7.9?

invinciblycool: Yeah you are right, I missed that `int()`.

src/pages/python/how-to-convert-strings-into-integers-in-python/index.md
invinciblycool: Yeah that would be better.

invinciblycool:
Updated as requested.

Bouncey:
Great :+1: 
## LGMT!

Thanks @invinciblycool :sparkles: :sparkles:

PR #272, opened 2017-09-14
Check for throttle/debouce echos before querying

### LINKED ISSUE #230: Flicker while search results load
No gif, because the flicker is probably too fast to properly capture it. Instead, I have
### Steps to reproduce
1. Search for something, like "Beau Carnes"
2. Add or remove one character

You'll notice that you immediately get the expected flicker of the skeleton and then to the actual results. This is expected. The problem happens right after that. It shows a flicker of the skeleton again, then back to the results. This double flicker is quite annoying and unexpected.

This problem is most noticeable on repeated searches with fast connection. If the search takes long enough, there will only be a single (expected) flicker.

### Possible causes
I'm not certain of the actual cause, but here's what I do know:
- It's not the event. As far as I can tell, `onChange` only fires once per change, and `onSubmit` doesn't influence anything.
- Commenting out line 21 in `searchEpic.js` did appear to solve the problem.
https://github.com/freeCodeCamp/guides/blob/27f2603c36383b0d06b90f628950a22d017cabdc/src/layouts/components/search/redux/searchEpic.js#L19-L22

Furthermore, we might want to completely disable the skeleton for repeated searches and only show it for the first one. It would eliminate the expected flicker.
It would show the skeleton from the first key press, and it would show a 'no results' message after a timeout if the user doesn't continue typing. If the first results have loaded and the user searches again while still on the search page, the skeleton isn't shown. This would probably be more comfortable than the constant flicker, and more consistent with other typeahead experiences.

Bouncey:
closes #230 

Check for throttle/debouce echos before querying

Also, if we have results, there is no need to display the skeleton again

### DIFF
--- a/src/LayoutComponents/search/redux/searchEpic.js
+++ b/src/LayoutComponents/search/redux/searchEpic.js
@@ -9,11 +9,11 @@ import {
 } from './';
 
 let previousSearchTerm = '';
-let Xms = 500;
 const requestUrl = 'https://freecodecamp.duckdns.org';
 const nullAction = { type: 'null' };
 
 function searchTermEpic(actions$, { getState }) {
+  const Xms = 500;
   const source$ = actions$
     .filter(({ type }) => type === types.updateSearchTerm);
 
@@ -23,13 +23,6 @@ function searchTermEpic(actions$, { getState }) {
     )
     .flatMap(() => {
       const { searchTerm } = getState().search;
-      // if user is pressing backspace
-      // increase debounce/throttle time
-      if (searchTerm.length < previousSearchTerm.length) {
-        Xms = 1000;
-      } else {
-        Xms = 500;
-      }
       // if the search term is over 2 chars and
       // this is not a throttle/debounce echo
       if (
--- a/src/LayoutComponents/search/redux/searchEpic.js
+++ b/src/LayoutComponents/search/redux/searchEpic.js
@@ -9,11 +9,11 @@ import {
 } from './';
 
 let previousSearchTerm = '';
+let Xms = 500;
 const requestUrl = 'https://freecodecamp.duckdns.org';
 const nullAction = { type: 'null' };
 
 function searchTermEpic(actions$, { getState }) {
-  let Xms = 500;
   const source$ = actions$
     .filter(({ type }) => type === types.updateSearchTerm);
 
--- a/src/LayoutComponents/search/redux/searchEpic.js
+++ b/src/LayoutComponents/search/redux/searchEpic.js
@@ -8,11 +8,12 @@ import {
   updateSearchResults
 } from './';
 
+let previousSearchTerm = '';
 const requestUrl = 'https://freecodecamp.duckdns.org';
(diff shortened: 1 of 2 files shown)


src/layouts/components/search/redux/searchEpic.js
```diff
@@ -22,9 +23,17 @@ function searchTermEpic(actions$, { getState }) {
     )
     .flatMap(() => {
       const { searchTerm } = getState().search;
+      /* if user is pressing backspace or we are
+      getting a debounce/throttle echo, do nothing */
+      if (searchTerm.length <= previousSearchTerm.length) {
```
systimotic: We can combine these two if statements to remove a bit of code duplication:
```js
if (searchTerm.length > 2 && searchTerm.length >= previousSearchTerm.length)
```

src/layouts/components/search/redux/searchEpic.js
```diff
@@ -22,9 +23,17 @@ function searchTermEpic(actions$, { getState }) {
...
+        previousSearchTerm = searchTerm.slice(0);
+        return Observable.of(nullAction);
+      }
       if (searchTerm.length > 2) {
+        previousSearchTerm = searchTerm.slice(0);
```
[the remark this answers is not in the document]
Bouncey: To be honest, I'm not comfortable with this as it is using something outside of `state` to effect the logic of ththe app. 

`previousSearchTerm = searchTerm` could make the app do weird things as one is only a reference to the other. 

`String.slice()` returns a new string, so using this to copy a string should not have any side effects.

src/LayoutComponents/search/redux/searchEpic.js
```diff
@@ -22,9 +23,23 @@ function searchTermEpic(actions$, { getState }) {
     )
     .flatMap(() => {
       const { searchTerm } = getState().search;
-      if (searchTerm.length > 2) {
+      // if user is pressing backspace
+      // increase debounce/throttle time
+      if (searchTerm.length < previousSearchTerm.length) {
```
systimotic: The throttle and debounce functions are set up on the first run on this function (page load) and are never run again. Therefore, changing the `Xms` doesn't have any effect.

Bouncey: Good catch!

Bouncey: I have moved Xms outside of the scope of the epic so it should work as expected now.

systimotic: @Bouncey I added `console.log('run')` to line 19 as a test. It only logged on page load, and not after it. That would mean the function is only fired once, so `Observable.merge` is also only called once, which in turn means that while the variable may change, because the function isn't called again with the new parameter the delays are never updated. `flatMap`'s callback is the only thing that is run repeatedly.

Bouncey: Ok 👍

Bouncey:
Sweet!

PR #283, opened 2017-09-18
Separate quote in machine learning for emphasis

erictleung:
I haven't tested this locally, but if Markdown blockquotes won't render correctly, I can change to be an HTML blockquote instead.


Bouncey:
## LGTM 👍 

thanks @erictleung ✨ ✨

PR #284, opened 2017-09-18
Fix capitalisation of "JavaScript" in index

systimotic:
I noticed the capitalisation of "JavaScript" in the index was off, so I fixed it 😄


Bouncey:

PR #279, opened 2017-09-17
Grey out article title if it is a stub

Bouncey:
As per suggestion by @evaristoc in gitter one day

screenshot from 2017-09-17 21-10-36

If the side nav is linking to a stubbed article, the title is greyed out

PR #285, opened 2017-09-18
Add a badge for stubs

systimotic:
From https://github.com/freeCodeCamp/guides/pull/279#issuecomment-330210215:
>I really like this change!
Idea: add a badge to stubbed articles to make it clear what it means when an article is grey.
Stub badge

If you don't like this (I could see how it might be too much) or want it to be different, let me know!

### DIFF
--- a/src/css/main.css
+++ b/src/css/main.css
@@ -200,8 +200,19 @@
   overflow-y: scroll;
   max-height: 90vh;
 }
+
 .stubbed {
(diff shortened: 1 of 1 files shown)


QuincyLarson:
@systimotic This is a great idea! Saves people a click. 

@Bouncey What do you think?

Bouncey:
## LGTM :+1: 

PR #304, opened 2017-09-22
Remove depreciated React.Proptypes call

Bouncey:


Bouncey:
React v16 is comming next week

PR #311, opened 2017-09-23
Fixed broken link and updated compile instruction

masquerade817:
The hyperlink to 'Hello World' program explanation in the Java documentation wasn't properly placed. Fixed it.

Also, a novice programmer might not be familiar with the use of terminal / Command Prompt and may not know that you need to be in the directory the file is contained in while compiling and running the code. Hence, made a slight change to the compile instruction for Hello World program.


Bouncey:
Great!

Thanks for the edit @masquerade817 

PR #324, opened 2017-09-23
Added c lenguage hello world

augur97:
Hi 
I have added a hello world example for C
I would appreciate if someone could look it over since English is not my native lenguage
Thanks in advance
Augur97


Bouncey:
Thanks for this @augur97 :+1: 

PR #318, opened 2017-09-23
Added Guide for Git Commit

JRFrazier:

PR #335, opened 2017-09-23
Created commit page

ZaninAndrea:
I wrote not only about the command `commit`, but also about what commits are. I'm not sure whether this explanation belongs here or the main page.

### DIFF
--- a/src/pages/git/git-commit/index.md
+++ b/src/pages/git/git-commit/index.md
@@ -2,14 +2,15 @@
 title: Git Commit
 ---
 ## Git Commit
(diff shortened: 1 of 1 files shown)


HKuz:
@ZaninAndrea - ack, I hate to be the bearer of bad news, but another pull request (#318) that filled in the Git Commit stub was submitted a few hours before this one, and just accepted 😅 . Looking at your changes, there's likely to be a lot of merge conflicts if you try to rebase. That said, I like how you explained what a commit is in your first paragraph and think it will add to the new article. 

What I suggest we do is close this PR, then you can grab the latest Git commit article, layer your commit explanation into it, then open a new PR.

Thanks so much for contributing, and don't be discouraged we're closing this - it's a bit weird that out of all the article stubs, two people worked on the same one within hours of each other.

PR #336, opened 2017-09-23
Linear Regression Explanations and Examples

kirtanechiron:
quick high level explanation with links and code to explore interactively


Bouncey:
Thanks for this article @kirtanechiron :+1:

PR #345, opened 2017-09-23
Add hdd article

crclark96:
add information about hdd architecture and usage

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@ package-lock.json
 navData.json
 .svn
 *.DS_Store
+*~
 
 # Logs
 logs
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -3,6 +3,7 @@ package-lock.json
 navData.json
 .svn
 *.DS_Store
+*~
```
Bouncey: What is this doing? Never seen it before is all

crclark96:
Ignores emacs autosave files

I guess I'm the only one here using emacs right now :P

Bouncey:
Sweet!

Thanks for the PR @crclark96 

PR #366, opened 2017-09-24
Fix link in JavaScript for-loop article

david-daly:
The for-loop article in the JavaScript section of the guide has a problem with the link where all of the text on the page is a link to the MDN article for for loops:

image

This pull request should fix that, adding the link to just the MDN link at the end of the article.


HKuz:
LGTM! 👨‍🎤 ✋ 🎤  Thanks for fixing this, @david-daly!

PR #369, opened 2017-09-24
Add basic info on Neural Nets

Sayan98:


Sayan98:
I have deliberately avoided adding any math. If need be, please ask.

HKuz:
This LGTM!! 🤔 💭 🤖 Thanks for the great write-up, @Sayan98! In terms of adding the mathematics, I believe we're still working on getting MathJax support to the Guide so we'll be able to write the formulas in Tex/LaTeX, but it's not in place yet. From a broader perspective (my opinion only), I say we focus on getting the general concepts and theory explained in as many articles as we can first, then we can go back and add the underlying math concepts as necessary 👍

Sayan98:
Okay !

PR #380, opened 2017-09-25
Fix typo

tuanhm2911:
Add a space between . and A

### DIFF
--- a/src/pages/javascript/callback-functions/index.md
+++ b/src/pages/javascript/callback-functions/index.md
@@ -9,7 +9,7 @@ The first thing we need to know is that in Javascript, functions are first-class
 
 ## Callback Functions
 
-A **callback function** is a function that is passed _as an argument_ to another function, to be "called back" at a later time. A function that accepts other functions as arguments is called a **higher-order function**, which contains the logic for _when_ the callback function gets executed. It's the combination of these two that allow us to extend our functionality.
+A **callback function** is a function that is passed _as an argument_ to another function, to be "called back" at a later time.  A function that accepts other functions as arguments is called a **higher-order function**, which contains the logic for _when_ the callback function gets executed. It's the combination of these two that allow us to extend our functionality.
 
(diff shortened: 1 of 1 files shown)


erictleung:
@tuanhm2911 thanks for the PR! Can you comment on why this space is needed? It doesn't seem necessary to me in this context.

texas2010:
oh ok. it makes sense. add space because of this.
screen shot 2017-09-25 at 4 03 42 pm


I found that on https://guide.freecodecamp.org/javascript/callback-functions

erictleung:
@tuanhm2911 @texas2010 ah I see. I don't think adding a space there will make a difference. To create an actual space, you'll need to add [a non-breaking space](https://stackoverflow.com/questions/13571706/how-to-add-white-spaces-in-html-paragraph), which I don't think will scale if we add this to all instances of this.

cc/ @Bouncey @systimotic second opinions?

systimotic:
Thanks for the PR @tuanhm2911! 

First of all, let's make sure I understand the issue this is trying to solve. When there is a space between a `.` and a `A`, it looks smaller than a regular space in this font.

As @erictleung mentioned, Markdown and HTML ignore double spaces.
```markdown
There   are   three   spaces   between   every   word   in   this   sentence,   but   you   will   only   see   one.
```
becomes:
There   are   three   spaces   between   every   word   in   this   sentence,   but   you   will   only   see   one.


I'm not sure if adding an extra space is the right solution here. I agree with Eric, if we try to fix this issue this way, we'll have to remember to that in all of the cases where this shows up this way. That would be difficult to keep up.

texas2010:
it seems we cant fix it. if we cant. we could close it.

erictleung:
Closing as a non-issue. 

@tuanhm2911 if you find this spacing issue in more than one place, you can raise an issue here to add some CSS to increase the letter spacing.

Happy coding!

PR #385, opened 2017-09-25
Added article for CSS Borders

chiragbhansali:


HKuz:
@chiragbhansali - thanks for making the changes! So you know, you can format in-line code with single backticks (\`border\`) - you don't need to use triple backticks 👍 . I'm having trouble doing the file move operation from my end to get rid of the duplicate article. What I suggest is we merge this, then feel free to open a new PR that moves the file into the properties folder (and replaces the stub in there). Thanks!

PR #418, opened 2017-09-25
JavaScript - Random Method: Update index.md

ChrisWcs:
Here is a start to the article on how to produce random numbers with Math.random which is what I assume the-random-method meant. Please let me know of any changes needed. Myself or another developer can always expand more on this subject in the future.


texas2010:
thanks!

PR #448, opened 2017-09-26
add information on imposter syndrome

svram:
added a description of what imposter syndrome is, what it looks like and how to tackle it


svram:
glad to contribute 👍

PR #452, opened 2017-09-26
Remove titles from search result descriptions

### LINKED ISSUE #289: Remove markdown symbols from search results
Currently search results show markdown like this:

We can probably remove the title from the text completely. It's possible an article could start with a table or other markdown syntax, but we can cross that bridge when we get there.

systimotic:
Closes #289 

Before:
20170926_12-12-08_firefox

After:
20170926_12-11-36_firefox

Maybe the links should be addressed too?

### DIFF
--- a/src/pageComponents/search/Results.jsx
+++ b/src/pageComponents/search/Results.jsx
@@ -13,7 +13,6 @@ import {
 } from '../../LayoutComponents/search/redux';
 
 const httpRE = /^https?/i;
-const articleMetaRE = /^\-\-\-[\w\W]+?\-\-\-/;
 
 const faNames = {
   challenge: 'free-code-camp',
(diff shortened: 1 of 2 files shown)


src/pageComponents/search/Results.jsx
```diff
@@ -59,6 +60,15 @@ MediaWrapper.propTypes = {
   url: PropTypes.string.isRequired
 };
 
+function formatGuide(text) {
```
systimotic: Is this the right place for this function?

Bouncey: I think it might be better to move this to the `searchEpic`, which mean is will only be done once for each result payload, instead of every render

Bouncey: What I was doing before was a bit naughty to be honest

systimotic:
@Bouncey Updated!

Bouncey:
## LGTM 👍 

QuincyLarson:
@systimotic This is a huge improvement to readability. Thanks for implementing this!

PR #459, opened 2017-09-26
Created and populated Javascript loops dir

johnkennedy9147:
Created loops dir
Moved all articles pertaining to loops to new dir
updated index.md in loops dir
deleted several stubs which already had articles


davidkwan95:
I have a suggestion to use something like `for...in` and `for...of` in the title so that it is more consistent to other documentations in the web.

johnkennedy9147:
is the for...in format common? 
I was mainly concerned with enhancing the tree structure to improve navigability, though I did amend two of the titles to be more consistent with the others.

davidkwan95:
a few popular sites like mozilla documentation and tutorialspoint use such formatting in the title
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in
https://www.tutorialspoint.com/javascript/javascript_forin_loop.htm

In the end it's just a matter of suggestion though

johnkennedy9147:
Rather than just loops I was thinking it might make sense to expand the dir to control flow and move in if and switch statements

johnkennedy9147:
@systimotic it just occurred to me that you might not be notified automatically that I made the requested changes plus a couple of other changes. I pushed the changes in a new commit, is there anything else I need to do?

PR #460, opened 2017-09-26
Python: Web Frameworks - index.md updated

BitYog:
Fixed grammar errors. Added a paragraph about Django and Flask.


texas2010:
thanks!

PR #466, opened 2017-09-26
First draft for DSU

physicsjoe:


texas2010:
@physicsjoe Please check that out. since it has conflicts in the file. more like someone did work on that. thanks

HKuz:
I took a crack at resolving these conflicts, this should be good to go. I did use the hyphenated form of "stand-up" everywhere, as it always seemed to be in adjective form referring to a "stand-up meeting", plus that's how Wikipedia uses it 😄

PR #365, opened 2017-09-24
Add article for JavaScript `typeof` keyword

david-daly:
Add some basic information about the JavaScipt `typeof` keyword, why it is useful and link to MDN Documentation fro more information.

PR #470, opened 2017-09-26
Initial article for JavaScript typeof

ksjazzguitar:
This is my first try at this, please let me know if I'm not doing this correctly

### DIFF
--- a/src/pages/javascript/typeof/index.md
+++ b/src/pages/javascript/typeof/index.md
@@ -9,34 +9,40 @@ This article gives a brief introduction to the concept and usage of the `typeof`
 The `typeof` operator allows you to find the data type of a variable or value. The type is returned as a *string*.
 
 This example shows how it can be used to find the type of various variables:
+
 ```javascript
 var name = "Carla";
+
 var age = 37;
+
 var dinner = {
   appetizer: "shrimp cocktail",
   entree: "ribeye steak, medium rare",
   dessert: "baked Alaska"
 };
+
 var hobbies = ['Sailing', 'Mountain Climbing', 'Coding'];
 
-function pickHobby(hobbyIndex) {
-  if (hobbyIndex<0 || hobbyIndex>hobbies.length)
-    return "I don't recognize that hobby.";
-  else
-    return "Let's do " + hobbies[x] + "!";
+function firstHobby() {
+    return hobbies[0];
 }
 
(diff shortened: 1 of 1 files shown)


IsaacAbrahamson:
I have a couple comments. Some of the code you have is not directly related to typeof, and could possibly add some confusion. For example, you define a somewhat complex function:
```javascript
function pickHobby(hobbyIndex) {
  if (hobbyIndex<0 || hobbyIndex>hobbies.length)
    return "I don't recognize that hobby.";
  else
    return "Let's do " + hobbies[x] + "!";
}
```
this has absolutely no bearing on typeof, as you just do this below:
```javascript
console.log("typeof pickhobby: " + typeof pickHobby);
// typeof pickhobby: function
```
What you have in the function doesn't change typeof, it could only possibly add confusion. Maybe just a basic function would be simpler?
```javascript
function pickHobby() {
    return hobbies[0];
}
```
My opinion for the console.logs. I think it might be good to add a space to reduce ambiguity. Just by looking at the comments, it is hard to tell right off bat whether the comment is for the log above it or below it. Maybe a line space between the logs would explicitly show which comment is for the log? Also, would it be better to simply log the typeof and not the string? I think making the guides as basic and simple as possible is a good thing. Thus
```javascript
console.log( typeof hobbies);
// object
```
would be preferable to
```javascript
console.log("typeof pickhobby: " + typeof pickHobby);
// typeof pickhobby: function
```
as it shows only typeof and nothing else to consider. What do you think?

david-daly:
@IsaacAbrahamson @ksjazzguitar I also have a pull request for adding a typeof article that has been open for 3 days #365 

Should we try to merge both of these articles, or just merge one or the other?

ksjazzguitar:
@david-daly I'm new to this and it sounds like you were first. Feel free to merge or not merge what you want from mine.

HKuz:
Hi everyone! Thanks for working together on this article. As @david-daly pointed out, there was another PR opened for it that I just merged. There's been a huge response from contributors since the announcement of the Guide, and unfortunately, a few collisions to edit the same article (typically, we try to favor the older PR when that happens). 

@ksjazzguitar - I'm going to close this PR given all the conflicts, but feel free to check out the article as it stands and open a new PR if you have any suggestions/edits! Thanks, all for contributing ✨ 🎉 🎊

PR #476, opened 2017-09-27
Add php example

RavinduThaveesha:


RavinduThaveesha:
Add Singleton PHP example

PR #479, opened 2017-09-27
added unary plus and parseInt

viktormelnychuk:
added content with unary plus and parseInt function


erictleung:
@viktormelnychuk LGTM! 👍 Thanks for taking the time to contribute! ✨

PR #429, opened 2017-09-25
Flesh out stub "Z Index"

IsaacAbrahamson:
Please review my guide to Z Index. Thank you!

PR #481, opened 2017-09-27
added article for z-index

chiragbhansali:


jamesjpk123:
Duplicate of #429, precedent of #519. We need to merge these commits together.

HKuz:
Hi @chiragbhansali - as @jamesjpk123 notes, this PR is a duplicate of one that was opened earlier 😢. I think all three PRs add good information, so what I suggest is you and @jamesjpk123 give any feedback to @IsaacAbrahamson over on #429. I definitely think the note about `z index` only working on positioned elements is worth bringing over 👍

PR #485, opened 2017-09-27
Border Property: Duplicate Property

chiragbhansali:


HKuz:
Thanks for getting rid of the duplicate article, @chiragbhansali - final step will be to move the article you wrote on Border Property into the properties folder 😆 👍

PR #495, opened 2017-09-27
Developer tools/npm

gadkadosh:
An article for npm.
Any ideas for additions/changes?
I tried to keep it as concise as possible, considering this is probably the single most important web dev tool. 
Too issues I encountered:
1. The font seems to cause --switch (dash-dash) to look like just a long dash. Seems confusing to me. A workaround would be to always have those in a code markdown block.
2. I believe it should be called npm rather than Npm. At least it seems to be consistently presented in that spelling.


dhcodes:
**LGTM**. Thanks @gadkadosh. I did change it to `npm` for you because I agree. It's the way the creators want it.

PR #511, opened 2017-09-28
remove explicit references to Ubuntu

rhainne:
Ubuntu is just one Linux distro. I think it should not get special treatment in the Linux index. Please consider it.
Specially because of this: 
https://www.gnu.org/philosophy/ubuntu-spyware.en.html
https://www.youtube.com/watch?v=CP8CNp-vksc

### DIFF
--- a/src/pages/linux/index.md
+++ b/src/pages/linux/index.md
@@ -5,6 +5,6 @@ title: Linux
 
 Linux is the most popular operating system in the world. It powers 90% of the world's servers, and is the basis of Google's popular Android operating system.
 
-There are also many versions of Linux designed to be run on personal computers.
+There are also many versions of Linux designed to be run on personal computers, such as Debian or Ubuntu
 
 For more facts about Linux, read Quincy Larson's <a href='https://medium.freecodecamp.org/linux-is-25-yay-lets-celebrate-with-25-rad-facts-about-linux-c8d8ac30076d' target='_blank' rel='nofollow'>"Linux is 25. Yay! Let’s celebrate with 25 stunning facts about Linux"</a>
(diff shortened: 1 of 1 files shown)


src/pages/linux/index.md
```diff
@@ -5,6 +5,6 @@ title: Linux
 
 Linux is the most popular operating system in the world. It powers 90% of the world's servers, and is the basis of Google's popular Android operating system.
 
-There are also versions of Linux designed to be run on personal computers, such as Ubuntu.
+There are also many versions of Linux designed to be run on personal computers.
 
-For more facts about Ubuntu, read Quincy Larson's <a href='https://medium.freecodecamp.org/linux-is-25-yay-lets-celebrate-with-25-rad-facts-about-linux-c8d8ac30076d' target='_blank' rel='nofollow'>"Linux is 25. Yay! Let’s celebrate with 25 stunning facts about Linux"</a>
\ No newline at end of file
+For more facts about Linux, read Quincy Larson's <a href='https://medium.freecodecamp.org/linux-is-25-yay-lets-celebrate-with-25-rad-facts-about-linux-c8d8ac30076d' target='_blank' rel='nofollow'>"Linux is 25. Yay! Let’s celebrate with 25 stunning facts about Linux"</a>
```
systimotic: I think this is a good change, this post isn't about Ubuntu specifically, so it shouldn't say so. 👍

src/pages/linux/index.md
```diff
@@ -5,6 +5,6 @@ title: Linux
 
 Linux is the most popular operating system in the world. It powers 90% of the world's servers, and is the basis of Google's popular Android operating system.
 
-There are also versions of Linux designed to be run on personal computers, such as Ubuntu.
+There are also many versions of Linux designed to be run on personal computers.
```
systimotic: I think having Ubuntu here is fair. It's not mentioning that it's good or bad, it's saying that Ubuntu is a version of Linux designed for personal computers. It's a good example because it's still the most popular one. I think adding `many` is a good change. Adding another example is fine with me too: "such as Ubuntu and Debian."

PR #522, opened 2017-09-29
TypeScript: Adding more detail of null-type and undefined-type

specterbn:


Bouncey:
Thanks for these articles @specterbn, sorry for the delay

PR #523, opened 2017-09-29
Reducing image size to fit the screen.

pedrosaisse:


systimotic:
Hi @pedrosaisse!
Thanks a lot for this PR.
The issue you're solving here is a recurring issue. I think the best way to solve it would be to use CSS to enforce a maximum width for all images in the content. I've opened #542 about this, and I'm closing this PR.

There is still a lot to be done on the Guide, and if you know how to make this change, you definitely know enough to do other things here too. I'd love to see more contributions from you!

Happy coding!

PR #541, opened 2017-09-29
Fix breadcrumb rendering on direct link

systimotic:
The breadcrumb would break when you linked to an article directly, rather than browsing to it:
20170929_15-01-26_firefox
Should be:
20170929_15-01-48_firefox
Example: https://guide.freecodecamp.org/html/attributes/

This PR fixes this issue. This was happening because on the live server, a trailing slash was added to each path, while there was none locally. I've switched to using a regex that grabs everything between the first slash and the last one (if there is one), making sure not to include the final slash in what it selects.

I do have one concern about this: my current version only grabs letters, numbers, slashes, and dashes. If there are other characters in the path, it will fail. I could also switch to a version that grabs _everything_, but I'm not sure if we'd want that either.

### DIFF
--- a/src/templateComponents/Breadcrumbs.jsx
+++ b/src/templateComponents/Breadcrumbs.jsx
@@ -12,8 +12,10 @@ function Breadcrumbs(props) {
   if (path === '/') {
     return null;
   }
-  // remove leading '/'
-  const pathMap = path.slice(1)
+
+  const pathMap = path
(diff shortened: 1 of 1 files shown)


Bouncey:
I heard you like regex in your regex

We should be OK checking for just those chars. Most other chars would break URL's any how

systimotic:
Yes, I need all the regex

PR #546, opened 2017-09-30
Description of an Epic added

davinderpalrehal:
Epic description and example


davinderpalrehal:
Just found out that I added incorrect information.

PR #576, opened 2017-10-01
Updated API info and created new 'HTTP requests' file

BitYog:


texas2010:
I would ask you fix it but the problem is you delete your repo.  so you could work on it again.  sorry. 

another problem is one file will cause a problem in the master if i merge PR.  because one file is empty and missing - in the folder name.

PR #595, opened 2017-10-01
RAM: Updated index with a few starter paragraphs

mozebdi:
Very basic starter paragraphs about what RAM is, what its role is in a modern PC, and why more RAM is useful.


texas2010:
sorry about it. someone did write this article.

PR #611, opened 2017-10-02
Added article about RAM

Hijerboa:
Explained its name, purpose, and provided other sources to read about it.


texas2010:
sorry about it. someone did write this article.

PR #621, opened 2017-10-02
README: Add info on adding links to articles

erictleung:
Fixed some formatting too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,6 +74,7 @@ We've written the following guide to writing Guide articles to help you get star
 - [Article title](#article-title)
 - [Modularity](#modularity)
 - [General writing tips](#general-writing-tips)
+- [Attribution](#attribution)
 - [Formatting example code](#formatting-example-code)
 - [Adding images to articles](#adding-images-to-articles)
 - [Adding links to articles](#adding-links-to-articles)
--- a/README.md
+++ b/README.md
@@ -76,6 +76,7 @@ We've written the following guide to writing Guide articles to help you get star
 - [General writing tips](#general-writing-tips)
 - [Formatting example code](#formatting-example-code)
 - [Adding images to articles](#adding-images-to-articles)
+- [Adding links to articles](#adding-links-to-articles)
 - [Proper nouns](#proper-nouns)
 - [Where to get help](#where-to-get-help)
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -180,6 +183,22 @@ Then you'd just need to reference them in your markdown file with this syntax:
...
+**Markdown**
+
+```
+[freeCodeCamp](https://www.freecodecamp.org/)
+```
+
+**HTML**
+
+```
+<a href="https://www.freecodecamp.org">freeCodeCamp</a>
```
systimotic: I'm not sure if we should recommend this. I saw the conversation on Gitter, but I'd like to add some information:
We have a script that transforms Markdown links into HTML links with `target='_blank'` and `rel='nofollow'`. If we say that people can add HTML links themselves, we won't be adding these attributes.

HKuz: One slightly-related note we may want to include in this section, too, is how to link to other Guide articles. I realized on QA that a few of the relative links I used in the Git section aren't working on the live site (PR to fix them is coming... 😄 ). This format definitely works:

`[Working link to Guide Git stash article](https://guide.freecodecamp.org/git/git-stash/)`

HKuz:
An unrelated thing I just noticed is in the README, under the "Modularity" section we reference an "Other Resources" section for additional links, etc. Looks like the stub format uses "More Information:" - may want to update the README reference for consistency. We could get the change in with this PR since it's open, or open it up for a first timer.

README.md
erictleung: @systimotic thanks for mentioning the script that does this change. Not knowing that a script is used to transform these links might be confusing to contributors. How about mentioning that there is a script that does this and you can definitely write links as Markdown, but if you add in HTML links, you have to add those attributes (the target blank and rel no follow) as well? I am a bit confused on this statement

> If we say that people can add HTML links themselves, we won't be adding these attributes.

@HKuz I'll wait on the decision on relative links to add them to the README 😄  I'm sure we'll get PRs to fix broken relative links in the future. Hopefully, we can find a more robust solution when the article structure changes.

erictleung:
@HKuz I can update the README to say "More Information:" to be consistent with the article subs themselves 👍

README.md
systimotic: @erictleung Sorry, that was very poorly worded. What I meant was that our script only processes Markdown links. It doesn't touch HTML links. If an HTML link does not have these attributes already, our script still won't add the attributes.

The mysterious script I'm talking about is `bin/normaliseArticles.js` and can be run with `yarn run normalise`. It is also used as a commit hook, meaning it runs before every commit you make locally.

I think updating the README to say that relative links should be like @HKuz showed would be a good idea. That way, it'll be easier to update everything once we get around to making relative links better (it's on my never-ending todo list 😄).

erictleung:
Changes made!

cc/ @systimotic @HKuz @texas2010

HKuz:
Thanks @erictleung - this LGTM!! I fixed a typo I found in one of the other sections 😄 - README is looking good 👍

PR #624, opened 2017-10-02
Delete empty TDD directory

systimotic:
This directory was empty and causing errors while building. It's preventing new updates from going live.
/cc @Bouncey


systimotic:
Thanks for getting to this so fast @Bouncey ❤️

PR #654, opened 2017-10-03
change to article

jello7654:
Remove instancse were writers' personal pref. seemed to be displayed in a too heavy handed way. Also removed portion disc. browsers , since by writers own admission he was unable to test Safari etc.


erictleung:
@jello7654 thanks for the PR! LGTM! 👍 Good editing on trying to keep objectivity in the articles.

PR #660, opened 2017-10-03
Fix build errors caused by frontmatter problems

systimotic:
_Another day, another broken build!_ 😂
_Yesterday, you got a free error, today you get three for the price of one!_

Three new issues that prevented the build from working. We should probably create some strict linting for the articles and add it to the CI to prevent things like this from happening so often.

/cc @Bouncey

### DIFF
--- a/src/pages/computer-science/interpreters/index.md
+++ b/src/pages/computer-science/interpreters/index.md
@@ -1,3 +1,7 @@
+---
+title: Interpreters
+---
+
 ## Interpreters
 
 Interpreters are a kind of translator. Interpreters take source code - code written in JavaScript, Python, and other languages. Then they translate it into code the computer understands.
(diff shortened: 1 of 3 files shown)


Bouncey:
@BerkeleyTrue Re: our conversation on Skype.

Bouncey:
Netfliy has a deploy preview bot that we are currently not using. The bot builds the PR and lets us know if it succeeds or fails, with a preview of the site on success.

We are trying to address this

Bouncey:
@texas2010 @HKuz @dhcodes @erictleung 

Please be aware that all articles need

```yml
---
title: Article Title for the UI
---
```

systimotic:
That would be really useful. I feel like I recall seeing that used here before, but it was removed, no?

Bouncey:
In the beginning yes, but I can't remember the reason it was taken away.

dhcodes:
Hmm. I think I just merged one last night where that was removed. I will unmerge, make the edit, and remerge.

PR #665, opened 2017-10-03
RAM: Updated index.md

BitYog:
Fixed typo


texas2010:
thank you for finding that. but someone did find it first and write an article.  thank again

PR #666, opened 2017-10-03
Euler's Method - Fixed broken image links

pranabendra:
1. Fixed broken image links (Links now contain the image file and not the page).

2. Replaced comma with semicolon


Bouncey:
If any PR deserves a `hacktoberfest` label, it it #666!

PR #674, opened 2017-10-03
Unbreak the build

Bouncey:
Someone merged an article that broke the build.

Sorry guys


systimotic:
:smile: Happens to the best of us! This proves once again that it would be useful to have some sort of automated test that warns us about this

Issue #670, opened 2017-10-03
Multiple Image Sections


nebrelbug:
Under HTML, there is:

Images
Elements/Images
Elements/Img Tag

systimotic:
Good catch @nebrelbug! Here are the links to the articles:
https://guide.freecodecamp.org/html/images/
https://guide.freecodecamp.org/html/elements/images/
https://guide.freecodecamp.org/html/elements/img-tag/

The first article is the most complete, but the last one has the best place and title. The contents of the first article should be added to the last one. The first two articles should then be removed.

nebrelbug:
I just found another one! There's also https://guide.freecodecamp.org/html/elements/html-img-tag.
Is https://guide.freecodecamp.org/html/elements/img-tag still best? Thanks!

nebrelbug:
Do my latest Pull Requests work?
Thanks so much!

systimotic:
Good catch once again @nebrelbug! Those PR's are a good start. I've opened a big PR that removes a lot of duplicate articles in the HTML section (#703), including the `html-img-tag` article, so I'm closing #680.

The others still need to be fixed, though. Can you make all the changes (including moving the content to the right article) on one branch and create a big PR that does the whole thing at once? You can do this using the GitHub interface, but you could also use [GitHub desktop](https://desktop.github.com/) to clone your fork and work on it on your own computer.

The reason I'm asking for a single PR is that otherwise we might get stuck with either no article that has the right content, or two articles with the same content 😅

To recap, the PR should:
- Add content from `html/images` to `html/elements/img-tag`
- Delete `html/images`
- Delete `html/elements/images`

siderio2:
Looks like this is affordable to me (almost totally newbie), so I would gladly try and submit a PR if you want.

systimotic:
Hi @siderio2! As @nebrelbug already started on this previously, we'll wait a little bit for their response on this particular issue. There are still plenty of other things to be done on the Guide though, your contribution there would be much appreciated! May I suggest https://github.com/freeCodeCamp/guides/issues/302, https://github.com/freeCodeCamp/guides/issues/319 or https://github.com/freeCodeCamp/guides/issues/367?

siderio2:
Sure, I'm already looking at those, and if @nebrelbug doesn't come around in the next days, I'll be ready to take over it.

systimotic:
@siderio2 Awesome, thanks!

nebrelbug:
[...]

PR #679, opened 2017-10-03
Delete because content should be at html/elements/img-tag

nebrelbug:
Delete because content should be at html/elements/img-tag
Issue #670


systimotic:
Closing as per issue discussion.

PR #688, opened 2017-10-04
Fix broken links & code in misc. merge sort guide

meowwwls:
The code examples were broken by partial links that were in the code, which also broke the links to the repl.it examples. 

The images remain broken, as I believe the file paths are invalid, as well as the fact that Markdown is not allowed inside of block-level elements.


systimotic:
Here's an article with working images: https://github.com/freeCodeCamp/guides/blob/master/src/pages/certificates/validate-us-telephone-numbers/index.md

Feel free to submit a new PR that fixes them for this article.

PR #690, opened 2017-10-04
add examples for python substring from string

svram:
added code samples and templates demonstrating slicing in python strings


pragadheesh:
Updated with some more information on the slicing arguments with PR https://github.com/freeCodeCamp/guides/pull/858

PR #703, opened 2017-10-04
Remove duplicate HTML articles

systimotic:
This PR removes the duplicate articles I found in the HTML section. There may be more, but I found these 30+ in my first quick look.


HKuz:
Well done, @systimotic 👏 ✋ 🙌  LGTM! I did notice two image articles and two comments articles still in there, but I'm going to merge this to try to avoid a situation where people edit two different stubs for the same topic. We can consider the section a work-in-progress 😄

systimotic:
🙌 Thanks for merging @HKuz! It'll always be a work-in-progress 😄 We already had #760 in progress to fix the image one. The comment one still needs to be fixed. @nebrelbug Are you interested in taking a look at that?

HKuz:
Touche, @systimotic - always a work in progress! 😄 I forgot about that image issue, good to see it's already fixed 👍

Issue #477, opened 2017-09-27
Use of "JavaScript" and articles in titles


kevinsmithwebdev:
I'm new here, but when I look at the list of JavaScript articles, several of their titles lead with "JavaScript...". This creates a confusing list as it seems to be out of alphabetical order, with entries starting with "j" sprinkled throughout. I would suggest not putting "JavaScript" in the title as it is redundant because we are in the JavaScript folder.

This may have also lead to some accidental duplication, for example there is an entry for "JavaScript for Loop" and "For Loops". Also there is "For in" and "JavaScript Forin Loop".

Additionally, I would suggest avoiding starting titles with articles (a, an, the), as in "The JavaScript Continue Statement" and "The Random Method". I would think that the key word of the article (that being defined) should be the first word in the title to create a more visually comprehensible list.

This is the list of titles as it stands now on the live site:

```
Advantages and Disadvantages of JavaScript
Angularjs Interview Questions
JavaScript Arguments
Arithmetic Operation JavaScript
JavaScript Assignment Operators
JavaScript Boolean
JavaScript Break Statement
JavaScript Callback Functions
Classes in JavaScript
JavaScript Closures
JavaScript Code Block
Code Linting in JavaScript
Comparison and Logical Operators
JavaScript Comparison Operators
JavaScript Loops
Conditional Ternary Operators
The JavaScript Continue Statement
Converting Strings to Numbers
JavaScript Dowhile Loop
Error Handling and Try Catch Throw
JavaScript Falsy Values
For in
JavaScript for Loop
For Loops
JavaScript for of Loop
JavaScript Forin Loop
Form Validation
Function Closures
Function Invocation
JavaScript Functions List
JavaScript Get Timestamp
JavaScript Global Variables
Higher Order Functions
HTML DOM getElementById Method
HTML Dom Innerhtml Property
If Else Statements
The JavaScript Ifelse Statement
JavaScript Immutable Types
JavaScript Isloggedinasync
Location Object
Location Reload Method
Loops and Iteration
Manipulate JavaScript Objects
Manipulating Cookies
Multiply Two Numbers with JavaScript
Naming Convention for JavaScript
Numbers
Onclick Event
Onload Event
Operators
Output
Popup Boxes
Promises
Prototypes
Regular Expressions Reference
The JavaScript Return Statement
Scopes in JavaScript
Semicolons in JavaScript
JavaScript Strict Mode
Switch Statements
JavaScript Ternary Operator
The JavaScript Global Object
[...]

PR #704, opened 2017-10-04
removed javascript from title #477

dhanushuUzumaki:


texas2010:
it is better to close it.  you need to redo this. sorry

PR #716, opened 2017-10-05
Added Tail Call Optimization Content

johnpaulada:


johnpaulada:
Hi @texas2010 ! I moved the Tail Call Optimization guide content here. Thanks! 😄

PR #724, opened 2017-10-05
add code samples on python string join

svram:
added code samples on the different ways to join string based iterables in python


johnkennedy9147:
@svram you need the title line to start `title: String Join Method`

svram:
added the title: line @johnkennedy9147

PR #725, opened 2017-10-05
add info on python string split

svram:
add code samples on python string splitting


johnkennedy9147:
@svram issue is the title line please change to `title: String Split Method`

PR #737, opened 2017-10-05
Added PHP page folder and index page

cjbagley:
I tried for a couple of hours to get the site to run locally so I could test the fact that adding a folder worked as I expected, however I just could not get it to work. If there is something else I needed to do, just let me know and I'll fix it!


texas2010:
@cjbagley  it seems you accidentally add .idea folder. please delete that. thanks!

cjbagley:
@texas2010 Whoops - my IDE created it and I didn't realise I'd not ignored it! Thanks for the spot :)

texas2010:
@cjbagley, welcome anytime!  that is why I always use git status before I add all file once. (in case I don't want to add all once by accident) Another way is to use `git add FILE_NAME`.

PR #738, opened 2017-10-05
Added on details in getting started with Bootstrap

justdroo:
**PR DESCRIPTION**
- Reflected the current documentation from v4 of Bootstrap to provide the basics of downloading and using Bootstrap within an application.
- Includes CDN and Package Manager methods
- Removed the comment about Bootstrap providing icons. This information should be reflected within Glyphicon components and does not have to do with getting started with Bootstrap.

**NOTES FOR REVIEWER**
- I don't explain much about what a CDN is but I wasn't sure if providing this information would help in getting Bootstrap set up. More than happy to include extra information if it would be useful.

### DIFF
--- a/src/pages/bootstrap/bootstrap-get-started/index.md
+++ b/src/pages/bootstrap/bootstrap-get-started/index.md
@@ -3,11 +3,62 @@ title: Bootstrap Get Started
 ---
 ## Bootstrap Get Started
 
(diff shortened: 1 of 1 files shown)


davcri:
Looks good to me.

@texas2010 which one is the duplicate article ?

texas2010:
@davcri yeah someone did write this article. that is why file have a conflict. so i have to close this.

PR #739, opened 2017-10-05
Improve bin scripts

systimotic:
Another refactor of the bin scripts. They both had the same page loop done in a slightly different way. Now they both use the same page loop as a util. https://github.com/systimotic/guides/commit/54816bed35af6dcba1a860472dab587576c8dd0e

I also added a formatted error for when there is a missing `index.md`. https://github.com/systimotic/guides/commit/b6d5b485114e5bebd4fe9d0686046bc6b698522f

The final change in this PR changes the link normalization. https://github.com/systimotic/guides/commit/d86f9f94d7f52abb1027cb2389ccb795c242f3d6 Doing this convinced me switching to a plugin is absolutely the right thing to do. 
The new regex catches some things that previously weren't caught (examples in meta and misc image gallery).
It also doesn't break by thinking there are links that are spread out across multiple lines, which should help with some of the issues where links removed content.
It still has issues with multiple links on the same line. The ternary operator helps with this if there are more then two, but not if there are exactly two. Switching to a plugin would help massively with this, as we could use the AST instead of trying to parse using regular expressions.

### DIFF
--- a/bin/createNavData.js
+++ b/bin/createNavData.js
@@ -1,7 +1,6 @@
 const fse = require('fs-extra');
-const { Observable } = require('rx');
 
-const { commonREs, info, readDir, pagesDir } = require('../utils');
+const { commonREs, info, readDir, loopPages } = require('../utils');
 
 const { isAStubRE, metaTitleRE } = commonREs;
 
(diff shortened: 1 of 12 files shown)


bin/normaliseArticles.js
```diff
@@ -39,27 +38,11 @@ This is a stub. [Help our community expand it](https://github.com/freecodecamp/g
...
-      const [ childText, url ] = formatted;
-      const anchor = (
-        `<a href='${url}' target='_blank' rel='nofollow'>${childText}</a>`
-      );
-      anchored = anchored.replace(str, anchor);
-    });
-  }
-
-  return anchored;
+  return content.replace(markdownLinkRE, (match, before, text, url) => {
```
Bouncey: `String.replace()` can take a function!?

Every day is a school day

bin/createNavData.js
```diff
@@ -1,80 +1,78 @@
...
 
 const { isAStubRE, metaTitleRE } = commonREs;
 
 const navData = {};
 
+function readIndex(path) {
+  try {
+    return fse.readFileSync(path, 'utf8');
+  } catch (err) {
+    throw `Could not read file at "${path}".`
```
Bouncey: Instead of throwing an error here, could me make the whole build process more stable by inserting a stub and trying again?

We could import `appendStub` from `normaliseArticles` to accomplish this, but we would need to write the file also so that Gatsby doesn't fall over.

Perhaps we could have an article CRUD module to do this for us.

I am open to suggestions, or you telling me it is a bad idea. I haven't had my coffee yet 👍

bin/normaliseArticles.js
```diff
@@ -39,27 +38,11 @@ This is a stub. [Help our community expand it](https://github.com/freecodecamp/g
...
+    return markdownLinkRE.test(text) ? match : (
```
Bouncey: I don't understand this bit?

```
if (there is a markdown link in the child text) {
  don't replace anything
} else {
  format the link
}
```

Have I got that right?

utils/loopPages.js
```diff
@@ -0,0 +1,16 @@
+const { Observable } = require('rx');
+
+const readDir = require('./readDir');
+
+module.exports = function loopPages(handler, dir = 'src/pages') {
+  return Observable.from(readDir(dir)).flatMap(articleParentDir => {
```
Bouncey: `SuperNit`

`.flatMap()` on a new line please and indent its function correctly 👍

bin/normaliseArticles.js
systimotic: 🏫 You can do all the fancy things now 🎩 😄

bin/normaliseArticles.js
systimotic: Yes, correct! It's a bit of a hack to try to help with the issue of having multiple links on one line.

Updating the regex to not allow `)` in the link destination would break for some links: https://en.wikipedia.org/wiki/Java_(programming_language)

If there are two links on a single line, this will still fail:
```
[link 1](http://example.com/1) text [link 2](http://example.com/2) text
```
`text` will equal `link 1](http://example.com/1) text [link 2`. I can't think of a way to probably catch this using regular expressions. Interestingly, I seems like this problem hasn't shown itself in the real world yet.

In the event that there are three or more links on one line (only [this closure article](https://github.com/freeCodeCamp/guides/blob/a36162f4ec524cc9ec7d20b1e11d4c3c2e4e276f/src/pages/clojure/lists-are-everything/index.md#when-to-use-a-list) has that), this approach will make it have the same result as without this change: the Markdown links are left untouched. The `regex.test` returns true, because the full text of the middle link is in the `text`.
The ternary operator is only for this case, three or more links on one line.

I'm sorry, I should've explained that more clearly in the original comment.

bin/createNavData.js
systimotic: I was going to say yes, but I've changed my mind, I'm going to tell you it's a bad idea 😄 ☕️

This formatted error is mainly meant for the CI. We could add the stub when we run into this, but that would just make the CI absorb the error. The actual Pull Request would still be broken.

Changing my mind again. This is run on build too, so that means this error would be resolved there too, making this less of an issue. But, to keep changing my mind, the stub would still be missing from the repo, which we don't want.

Final verdict: we want to throw an error to make sure the stub is added to the repo.

utils/loopPages.js
systimotic: @Bouncey 🙌 I like nits, they mean you care about the quality of the code. Updated!

BerkeleyTrue:
Deploy preview failed.

Built with commit c5d5d3cd3bea5b8076b50bd487ddab1c24f30499

https://app.netlify.com/sites/guide/deploys/5a2fb3faa1147720b60b72d1

QuincyLarson:
@systimotic Closing this since the codebase has changed so dramatically in the past year. Once we move the guide into the freecodecamp/freecodecamp we'd welcome your help with scripts for it.

PR #742, opened 2017-10-06
add information on social engineering

svram:
added definition, info and links on social engineering hack


dhcodes:
I made some small edits but **LGTM**. Thanks for contributing @svram!

PR #746, opened 2017-10-06
Add Python string join method page

richyvk:


texas2010:
Sorry, I close this because someone did write this article. thanks.

PR #755, opened 2017-10-06
planning poker

Omerdogan3:


HKuz:
This article LGTM! ♠️ ❤️ ♦️ ♣️  I just removed the stub language so the article would show as a link in the menu. Thanks for contributing, @Omerdogan3 !

PR #767, opened 2017-10-07
Add quick sort

Sayan98:


QuincyLarson:
@Sayan98 Thanks for writing this excellent article on Quick Sort. This is a great start, and I'm going to go ahead and merge this. 

The wikipedia articles related to these sorting algorithms often include excellent animated illustrations, and since these are creative-commons licensed, we can include (and cite) them as well, as I think they really give perspective into how these sorting algorithms work.

You may be able to find ways to improve and simplify this article, to make it more accessible going forward. I would welcome future improvements to this article from you, and other articles on algorithms like this one.

PR #773, opened 2017-10-07
adding chrome-firefox devtools article

siderio2:
Adding a stub page for Chrome-Firefox devtools


siderio2:
Just noticed one thing: in the navigation bar, now alphabetic order is broken, the page I've just added is now the first one, before 'Bower'. How can this be fixed, please??

texas2010:
please remove this file. I noticed you added this. src/pages/.index.jsx.swp on this PR

siderio2:
@texas2010 thanks for the advice, the file has been removed, but how was it created in the first commit?? I don't even know where it came from???

HKuz:
LGTM!! 🔨 🔧 💻 Thanks for adding this stub, @siderio2! As for the new file - not sure where that came from - we have a commit-script that runs when you commit your changes, but it shouldn't create new files. Only other thing I can think of is if you were working in an IDE or have a plugin that went rogue 🤔?

PR #774, opened 2017-10-07
Converting Between Radians and Degrees: added new article

robnixon:
completed the converting between radians and degrees article


HKuz:
LGTM!! ↙️ ▶️ ↘️  Thanks for contributing, @robnixon!

PR #775, opened 2017-10-07
Making accessibility top level topic

siderio2:
Moving Accessibility from Miscellaneous directory to /


siderio2:
Sorry, I've made a new commit adding a copyright line. Acording to https://www.w3.org/Consortium/Legal/2015/doc-license, this should be enough.

HKuz:
This LGTM!! Thanks for checking that out, @siderio2 and for your patience as we work through these PRs! 👍 🌶 ♨️

PR #776, opened 2017-10-07
Add clearing the text in the search bar on the leaving the search page.

### LINKED ISSUE #232: Reset search bar
Using the side nav after searching should clear the search bar input

marekdano:
I found the way how to clear the term in the search bar on the leaving the search bar. I call the resetSearch method to set the searchTerm to empty string. The call is in the componentWillUnmount method in the pages/search component.
closes #232

### DIFF
--- a/src/pages/search.jsx
+++ b/src/pages/search.jsx
@@ -16,7 +16,7 @@ import {
 const propTypes = {
   isSearching: PropTypes.bool,
   lastPage: PropTypes.string,
-  resetSearch: PropTypes.string,
+  resetSearch: PropTypes.func,
   results: PropTypes.arrayOf(PropTypes.object),
   searchTerm: PropTypes.string
 };
--- a/src/pages/search.jsx
+++ b/src/pages/search.jsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { PureComponent } from 'react';
 import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
 import { bindActionCreators } from 'redux';
@@ -16,6 +16,7 @@ import {
 const propTypes = {
   isSearching: PropTypes.bool,
   lastPage: PropTypes.string,
+  resetSearch: PropTypes.string,
   results: PropTypes.arrayOf(PropTypes.object),
   searchTerm: PropTypes.string
 };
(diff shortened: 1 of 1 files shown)


src/pages/search.jsx
```diff
@@ -16,6 +16,7 @@ import {
 const propTypes = {
   isSearching: PropTypes.bool,
   lastPage: PropTypes.string,
+  resetSearch: PropTypes.func,
```
Bouncey: `PropTypes.func.isRequired`

If the component try's to unMount and this function is not available, the app will crash

src/pages/search.jsx
```diff
@@ -42,27 +43,38 @@ function shouldShowResults(conditions) {
...
-  ) {
-    return <NoSupport />;
+class SearchPage extends PureComponent {
+  constructor() {
+    super();
+  }
+
+  componentWillUnmount() {
+    const { resetSearch } = this.props;
+    resetSearch('');
```
Bouncey: This action expects no arguments, you are creating an empty string for no reason

src/pages/search.jsx
```diff
@@ -42,27 +43,38 @@ function shouldShowResults(conditions) {
...
+    return (
+      <div>
+        <Helmet>
+          <title>Search | freeCodeCamp Guide</title>
+        </Helmet>
+        <h2 className='colourDarkGrey'>Search Results</h2>
+        {
+          (isSearching && !results.length) ?
+            <ResultsSkeleton /> :
+            shouldShowResults({ results, lastPage, searchTerm })
```
Bouncey: Now this page is a `class`, we should make this function a method on the class. By doing so, we will not need to pass props to it manually, as they will be available via `this.props`

src/pages/search.jsx
marekdano: Ok, I modified it.

src/pages/search.jsx
marekdano: Good, I'll remove the empty string.

src/pages/search.jsx
marekdano: Ok, I understand your comment. Just one question. What is the best practice? Moving the method "shouldShowResults" before "componentWillUnmount" or after.

Bouncey: I think we should go with 
```jsx
class Search extends PureComponent {
  // constructor

  // component lifestyle hooks

  // component methods

  // render() {...}
}
```

Thanks for asking @marekdano

marekdano: @Bouncey Thanks for your answer. I've just pushed my code. Hopefully it's good ;)

marekdano:
Let me know if you think that the code should be improved.

Bouncey:
Great Job @marekdano 

Sorry for the delay and thank you for your patience!

PR #777, opened 2017-10-07
add useful commands to vim page

carlosvigil:
will add more in the spirit of 'good enough'


QuincyLarson:
@carlosvigil This looks good. This is well written. Thanks for your contribution!

PR #781, opened 2017-10-07
Added article on Wireframing

sage-godspeed:
Also included a link for additional resources in the article


QuincyLarson:
@sage-godspeed Thanks for your contribution! This looks solid! Congratulations on your first accepted pull request here! We would welcome more articles on these sorts of design concepts from you.

PR #793, opened 2017-10-08
Adds vim modes page

shtanton:
My best explanation of how modes work in vim, I know there is some explanation in the useful parts page so I've gone into more detail here

### DIFF
--- a/src/pages/vim/modes/index.md
+++ b/src/pages/vim/modes/index.md
@@ -1,3 +1,7 @@
+---
+title: Modes in Vim
+---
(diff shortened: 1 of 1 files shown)


QuincyLarson:
@shtanton This is a solid article. Thank you for contributing. Also, congratulations on your first accepted pull request to this repo!

carlosvigil:
@shtanton Maybe we could put the useful commands bit into the introduction page as it's more of a crash landed type of tip/tone

PR #800, opened 2017-10-08
Added img-src guide

ApCoder123:
I added a guide about the img `src` attribute to replace this stub: https://guide.freecodecamp.org/html/attributes/img-src-attribute


HKuz:
This LGTM! 😎 📸 🖼 Thanks for contributing, @ApCoder123!

PR #806, opened 2017-10-09
Remove lightbox markup imported from forum

systimotic:
The forum has lightboxes, where you can click on an image to view it at a larger size. I think this wasn't accounted for in the importing process. I went through and did my best to fix the ones that I could find.

Related: https://github.com/freeCodeCamp/freeCodeCamp/issues/15918
/cc @nsuchy

### DIFF
--- a/src/pages/certificates/arguments-optional/index.md
+++ b/src/pages/certificates/arguments-optional/index.md
@@ -40,7 +40,7 @@ In the case that only one argument was passed, do not worry about how to prompt 
 
 ## Spoiler Alert!
 
-![687474703a2f2f7777772e796f75726472756d2e636f6d2f796f75726472756d2f696d616765732f323030372f31302f31302f7265645f7761726e696e675f7369676e5f322e676966.gif](//discourse-user-assets.s3.amazonaws.com/original/2X/2/2d6c412a50797771301e7ceabd554cef4edcd74d.gif)
+![warning sign](//discourse-user-assets.s3.amazonaws.com/original/2X/2/2d6c412a50797771301e7ceabd554cef4edcd74d.gif)
 
 **Solution ahead!**
 
--- a/src/pages/certificates/binary-agents/index.md
+++ b/src/pages/certificates/binary-agents/index.md
@@ -34,7 +34,7 @@ Make sure that each time you transcode a character from binary to decimal, you r
 
 ## Spoiler Alert!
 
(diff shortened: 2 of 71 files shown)


QuincyLarson:
@systimotic From looking at the files changed, it looks like you may have also changed some other lines of code unintentionally. Please take another look at this. Thanks for your help with this.

systimotic:
@QuincyLarson I forgot to mention in the original description that I also changed all instances of `6874747(...)e676966.gif` to `warning sign` for the following image:
warning sign
I came across it and thought it would be good to give it an actual descriptive title instead of that random key. It was an easy fix, so I went ahead and did it.

If I saw a broken code block on a page I removed a lightbox from, I also fixed that.

The last line of every edited file is marked as changed, but there is no change. I have a feeling it is caused by the [.gitattributes file](https://github.com/freeCodeCamp/guides/blob/177b88a6b0ebd42ae25fa3ac6f4eaf21164547bd/.gitattributes). If that's the case, it's a trade-off that's well worth it (without it, every single stub is marked as changed after every commit on Windows).

That was all I could find. Anything else I missed?

systimotic:
@QuincyLarson @nsuchy Could you take another look at this?

nsuchy:
These changes look solid to me. I notified many of the forum links are using http:// and the .com link. Could you update them to use https and the .org link? Some of the images do not define the protocol. FreeCodeCamp redirects users to the https version so I think all images should use HTTPS. Thoughts?

QuincyLarson:
@systimotic These look great! I agree with @nsuchy that going through and changing the links to https and freecodecamp.org would be a good step. Then we can merge your pull request.

Bouncey:
@systimotic fancy a `rebase` on this?

systimotic:
@Bouncey ✔️ Rebased!
I'll save the https and .org for another PR to make QA a little bit easier.

QuincyLarson:
@systimotic Awesome! Thanks for making these edits! 

@nsuchy Could you QA these real quick, and if you think they're ready, merge them?

nsuchy:
@QuincyLarson QAing now. I'll report back shortly.

systimotic:
Thanks @nsuchy!

QuincyLarson:
@systimotic Great progress here! 

@nsuchy Thanks for helping QA this!

PR #818, opened 2017-10-10
Ethical Breaches: Fixed "more information" links

arcdev1:
Looks like the post-commit action mangled the links at the bottom. Added proper markdown and line breaks.

### DIFF
--- a/src/pages/developer-ethics/ethical-breaches/index.md
+++ b/src/pages/developer-ethics/ethical-breaches/index.md
@@ -13,6 +13,6 @@ There have been a number of well publicized cases in which software was used to 
 
 --
 #### More Information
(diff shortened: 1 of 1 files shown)


davcri:
@zk433 thanks for reviewing ;) Square brackets are correct, because they are used to create a [markdown link](https://daringfireball.net/projects/markdown/syntax#link).

The problem is that the link isn't generated because there is a space between the square brackets and parenthesis:
```markdown
[correct link](https://www.freecodecamp.org/)

[bad link] (https://www.freecodecamp.org/)
```

arcdev1:
Good catch. Sent a new pull request.

davcri:
Tag me on the new PR so I'll merge it immediatly ;)

PR #832, opened 2017-10-11
Added article for Python list's sort method

pragadheesh:


invinciblycool:
Nice article @pragadheesh .

PR #876, opened 2017-10-13
HTML forms: Update index.md

sadeeqaji:


sadeeqaji:
I created new form with an input value

PR #878, opened 2017-10-13
Adds the vim copy and paste article

shtanton:
The articles details how to use copying/yanking, cutting and pasting in
vim, as well as registers.


MrAsrani1312:
Awesome. Thanks for your contribution 😄

PR #918, opened 2017-10-16
Added section on Parts of a compiler

neeilan:
Added a small section on the basic parts of a compiler (lexing, parsing and codegen).


QuincyLarson:
@neeilan Awesome! This article is off to a great start. We welcome additional sections that go more in-depth on these topics, when you have time to write them. And any other computer science concepts you'd be interested in covering. Thanks again for your help with this!

PR #927, opened 2017-10-16
Added article on jQuery plugins

wjbeshore:
Added article detailing jQuery plugins and some resources.


texas2010:
okay, I am confused about this pr. since you plan to add node article and jquery but you delete node article.

texas2010:
oh, i see a problem is. you have two PR but it is the same thing.....  this PR will stay open and you fix this.

texas2010:
i am going to close this. you can fix it again on new PR. thanks

PR #928, opened 2017-10-16
Bootstrap Index page: Fix invisible script tags

clottman:
These script tags appear as display:none in the rendered version right now. Adding the script type to the markdown makes them render as visible. 

I confirmed this running the guide site locally.


QuincyLarson:
@clottman Awesome! Thanks for figuring this out and fixing this!

PR #958, opened 2017-10-17
Created a small summary of C operators

angeliand:


QuincyLarson:
@angeliand Awesome! Any additional details you can add in the future will be helpful. This seems like a great start, though.

PR #959, opened 2017-10-18
Changed typo in Tail Call Optimization Guide and added clarification

johnpaulada:
I changed ECMAScript 2015 to ES5 because someone got confused after reading it. 😅

### DIFF
--- a/src/pages/software-engineering/tco-tail-call-optimization/index.md
+++ b/src/pages/software-engineering/tco-tail-call-optimization/index.md
@@ -33,7 +33,6 @@ In the previous example, the multiplication operation is executed last in the `r
 
 #### Example
 Here is an example of a JavaScript (ES5) factorial function using recursion **with** TCO.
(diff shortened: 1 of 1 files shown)


QuincyLarson:
LGTM.

QuincyLarson:
@johnpaulada just to make sure I'm not confused, ES5 is the really old version of JavaScript, and ES2015 is the same thing as ES6. This is what you meant here, right? Pre-ES6 JavaScript?

johnpaulada:
Yes sir. The code is in ES5 form but if run in a browser that supports ES6, Tail Call Optimization would kick in. Else, the result would be the same with the previous code.

johnpaulada:
By the way, is it okay to add trampolining to this guide or do I make another guide? It's a method for eliminating the stack overflow problem without relying on modern browser support.

QuincyLarson:
@johnpaulada when in doubt, make an additional guide article, then you can link to it from this article :)

johnpaulada:
Alright. I'll make a separate article then. Thank you sir @QuincyLarson ! 😄

PR #977, opened 2017-10-18
updated the definition and added one more example

jaitjacob:
1.) "element" is more suitable that the word "tool". 
2.) multi-line comment example has been added.


MrAsrani1312:
Awesome 👍 Thank you for your contributions 😃

PR #979, opened 2017-10-18
Added a missing bracket around the Source:

storrico87:


texas2010:
thank you for find incorrect. btw that link was broken. so I fixed it.  that is why ) is missing in the URL.

PR #985, opened 2017-10-18
JS/Booleans Added explanation of primitive bools

forkerino:
added some simple explanation and example for booleans when using JS. Added parts on type coercion and using falsy/truthy values in combination with || and &&


MrAsrani1312:
Awesome 👍 Thank you for your contribution 😄

PR #1086, opened 2017-10-19
changed Apple OS X to Apple macOS

jackHedaya:


texas2010:
someone did fix it. thanks

PR #1201, opened 2017-10-19
Updated content from stub

adamayd:


adamayd:
removed PR to resolve conflict.  Other accepted PR was much more in depth.

PR #1202, opened 2017-10-19
Update index.md with HTML version history

an732001:


texas2010:
Sorry someone did add HTML history at there.

PR #1209, opened 2017-10-19
Update String guide

ctam91:
Edited document for clarity.


davcri:
@ctam91 thank you for your contribute! Smoothing raw edges in existing articles is very important!

PR #1228, opened 2017-10-19
Edited "Promises | MDN" link

taruntejae:


texas2010:
Sorry, i cant accept this PR because markdown link will convert to HTML link in the auto script. that will be the same thing later.   but thank you!

PR #1251, opened 2017-10-19
Added about Personas

NikhilManapure:


davcri:
Hi @NikhilManapure you plagiarized parts of this article. This goes against our [Academic Honesty policy](https://www.freecodecamp.org/academic-honesty) so I cannot accept it.

- https://uxmastery.com/create-ux-personas/
- https://www.interaction-design.org/literature/article/personas-why-and-how-you-should-use-them
- https://www.usability.gov/how-to-and-tools/methods/personas.html
- http://www.uxbooth.com/articles/creating-personas/

If you think that all the content can be safely taken from the following sources, we still need to cite them properly.

Feel free to ask any questions.

PR #1234, opened 2017-10-19
Add a skeleton for Python Web frameworks

tarasmatsyk:

PR #1258, opened 2017-10-19
Add csharp for loop example

tarasmatsyk:


davcri:
Same as here https://github.com/freeCodeCamp/guides/pull/1234

PR #1266, opened 2017-10-19
Add definition

GennadiyFurduy:


dhcodes:
Waiting on https://github.com/freeCodeCamp/guides/issues/2503

jp-sauve:
Can this project's citation requirements override techterms.com's citation demands?

fvcproductions:
@jp-sauve I believe this is still in discussion but since most of the content for this PR is copied, it should be marked as `invalid` and closed. 

Even if there was a citation, I don't think the goal of the `guides` website is to be full of copy and paste paragraphs.

PR #1287, opened 2017-10-19
Add basic information with example

loki-one:
Add basic information regarding Window Location


loki-one:
Hi why did it fail? can someone explain

PR #1301, opened 2017-10-19
Removed spaces between word and colon

joe-gallant:


zk433:
Thank you @joe-gallant , merged.

PR #1311, opened 2017-10-19
Added link to w3schools 'display' information

SpinyUK:


texas2010:
could you put the link in the bottom?  not that way. thanks

src/pages/html/elements/span-tag/index.md
```diff
@@ -26,4 +26,4 @@ The code for the paragraph with red text looks like this:
 ```
 
 #### Differences between `span` and `div`
-The main difference is that `span` is an inline element, while `div` is a block element. This means that a `span` can appear within a sentence or paragraph (as in the example above), while a `div` will start a new line of content. Note that the CSS `display` property can change this default behavior, but that's way beyond the scope of this article!
+The main difference is that `span` is an inline element, while `div` is a block element. This means that a `span` can appear within a sentence or paragraph (as in the example above), while a `div` will start a new line of content. Note that the CSS <a href="https://www.w3schools.com/cssref/pr_class_display.asp">`display`</a> property can change this default behavior, but that's way beyond the scope of this article!
```
Bouncey: Just expanding on what @texas2010  said.

How about we link to [our article on the display property](https://guide.freecodecamp.org/css/css-display)

PR #1323, opened 2017-10-19
Clarified alt attribute

SpinyUK:


zk433:
@SpinyUK , thank you for your PR.
[According to W3C](https://www.w3.org/WAI/tutorials/images/decorative/), alt tag should NOT be provided for decorative images. If possible, decorative images should be provided using CSS instead.

SpinyUK:
as above, this change is no longer part of spec.

PR #1334, opened 2017-10-19
Create angular.md

malevolentninja:
Added Angular resources page.


Bouncey:
```sh
Error: ENOENT: no such file or directory, open 'src/pages/angular/index.md'
```

The parent directory, in this case `src/pages/angular` also requires an `index.md`. This can be added by running `yarn run normalise` and committing the changes

PR #1339, opened 2017-10-19
Add Changes to B Tag HTML Element

loki-one:
Basic information with example regarding B tag and a special Note when should one use it.


Bouncey:
Style and formatting edits applied

## LGTM :+1: 
Thanks @loki-one

PR #1342, opened 2017-10-19
Add instruction to launch "vimtutor" from command line

questionmarcus:
"vimtutor" is the de-facto "Vim 101". It is launched from the command line and can be run in different languages by using a 2 character country code.

I removed the previous section on launching using `Tutor` as a normal mode command as this does not work for me (Vim 8.0.1-1176)


Bouncey:
@questionmarcus it looks like we have a conflict. Can you checkout the latest changes to the article and see if your edits are still relevant please?

Many thanks

PR #1348, opened 2017-10-19
Added dnf and zypper commands

hemor:


Bouncey:
## LGTM :+1: 

Thanks @hemor

PR #1370, opened 2017-10-19
Updated an example of Supervised learning

askrocks:
Have added an example of Supervised Learning - Trying to differentiate between male and female images.


davcri:
**LGTM**!
Thank you for your contribute!

PR #1376, opened 2017-10-19
Added an example

frieddumpling:
Added "* Infiltrating: impersonating someone legitimate in order to gain physical access to a building/office etc., e.g. the coffee machine repair person" as an example


Bouncey:
Thanks for the edit @frieddumpling  

## LGMT 👍

PR #1403, opened 2017-10-19
Added a note in line 16

devopscodes:
Explained that the position in an array starts at 0 as opposed to 1.


IsaacAbrahamson:
LGTM. Thanks for contributing!

PR #1407, opened 2017-10-19
Provided examples of names, rgb, and hex

BenThom21:


Bouncey:
Thanks for the article @BenThom21 but it seems that this article has already been added. Feel free to add any edits to the current article

PR #1416, opened 2017-10-19
Added content

DimitrisZx:
Added descriptions for color names, rgb, hex values and why it is a better idea to use the last two.


ChaituVR:
Thank you for your Pull Request :+1:

PR #1435, opened 2017-10-19
Update index.md

bluecake:
Link to Medium article about colour theory


Bouncey:
@bluecake Thanks for the edit

Unfortunately we need something a bit more substantial. If I accept this PR, your link will be overwritten automatically when we run one of our chore scripts to format links and what-not. Please feel free to raise another PR with an article that can replace the stub. 

Many thanks

PR #1448, opened 2017-10-19
Added information about Data Packets in brief

JIBIN-P:
There may be some mistakes in the markup but the information is correct and I hope someone reviews it and if any extra information needs to be covered please do tell I will add it.


Bouncey:
Great article, Thanks @JIBIN-P :+1:

JIBIN-P:
Thank you

PR #1449, opened 2017-10-19
Added examples using the while statement.

ronaldotd:
"while", "break" and "continue" statements explained.


Bouncey:
@ronaldotd It seems that we have already have changes submitted for this article. If you feel you could add more, please either re-open this PR or create a new one

Many thanks

PR #1464, opened 2017-10-19
Update index.md

krish-ram:
Cloud Development Details


Bouncey:
@krish-ram We can't simply copy/paste content from proprietary sources

PR #1475, opened 2017-10-19
Fix grammar

weihungchin:


Bouncey:
You're changes do not change the article in any meaning full way.

Indentation is comparable to the three back ticks

PR #1483, opened 2017-10-19
added section 'tmux'

theaquacadet:


Bouncey:
Thanks for the edit @theaquacadet :+1:

PR #1493, opened 2017-10-19
Add CSS Transitions

thomlom:
Hi!

Here is my guide on CSS transitions. 
It may not be perfect or complete but I think it's enough for anyone who have worked a little bit with CSS to get started on transitions 😄

### DIFF
--- a/src/pages/css/css3-transitions/index.md
+++ b/src/pages/css/css3-transitions/index.md
@@ -3,13 +3,124 @@ title: CSS3 Transitions
 ---
 ## CSS3 Transitions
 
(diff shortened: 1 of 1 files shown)


src/pages/css/css3-transitions/index.md
```diff
@@ -3,13 +3,124 @@ title: CSS3 Transitions
...
+* `ease-in-out`
+* `cubic-bezier(n, n, n, n)`
+
+For example :
+
+```css
+transition-timing-function: linear
+```
 
+N.B : All the values above are in fact specifics `cubic-bezier`. `linear`, for instance, is equivalent to `cubic-bezier(0.25,0.1,0.25,1)`
```
ChaituVR: N.B??

thomlom: N.B means : "Nota Bene". It means here "Take careful notice of the following"

Bouncey:
Great article @thomlom :+1:

PR #1498, opened 2017-10-19
Add a simpler example and add example title

weihungchin:
Add a simpler example to illustrates the concept of closures


MrAsrani1312:
Awesome 👍 Thank you for the contribution 😄

PR #1500, opened 2017-10-19
Article fulfillment on "What is an API"

Josephvh:


texas2010:
Sorry, someone did write this article.

PR #1533, opened 2017-10-19
Csharp hello world

Xiryl:
Csharp hello world


Bouncey:
Thanks for the article @Xiryl

PR #1563, opened 2017-10-19
Labels can be used too!

weiffert:
Labels can be used for continue statements too.


QuincyLarson:
@weiffert Thanks for your contribution.

PR #1570, opened 2017-10-19
Update index.md

Rockleeland:
Added content to HTML entities


Bouncey:
@Rockleeland I need you to pad this article out

- What is a HTML Entity?
- Why should I use them?
- Where can I find a list of them?

Once you have added the suggested content, please re-open this PR

PR #1571, opened 2017-10-19
Update index.md

connorgallopo:
Added example of an IDE


Bouncey:
I would say that Atom is just a text editor

PR #1582, opened 2017-10-19
add String#index method with examples

robert-laws:
Add String#index method to the ruby string methods file.


Bouncey:
Thanks for the edit @robert-laws

PR #1597, opened 2017-10-19
Create a new folder in CSS for CSS preprocessors

StraubJones:
LESS and SASS are core to a lot of modern web development and key things for new developers to try and be familiar with. Short guides would be super helpful! But before diving into specific preprocessors, giving new learners a sense of what a CSS preprocessor is and what it does is really important.


Bouncey:
Thanks for the stub @StraubJones

PR #1599, opened 2017-10-19
Added JSON-Array article guide

hisivasankar:
Summary:
    Edited the existing index.md and added more content
about JSON array with an example


hisivasankar:
I update navData.json but it doesn't appear in changes. Hence I sent only my changes in index.md

IsaacAbrahamson:
LGTM. Thanks for contributing!

PR #1613, opened 2017-10-19
Added examples and use case

johnpyp:
Added more examples and explanation of filter as well as a use case to use it over a standard for loop.


ChaituVR:
Thanks for the Contribution. :+1: :100: :taco:

PR #1618, opened 2017-10-19
Added Description and some roles

b34s7m0d3:


Bouncey:
Thanks for the edits @b34s7m0d3

PR #1628, opened 2017-10-19
Add "pred" method documentation

yondri:


Bouncey:
Thanks for the edit @yondri

PR #1648, opened 2017-10-19
Update XAML

DannyDanielsan:
A small Textblock Xaml example.


Bouncey:
Thanks for the article @DannyDanielsan

PR #1658, opened 2017-10-19
Update info and examples to correct syntax on Javascript - Array.Prot…

artarmstrong:
…otype.pop


Bouncey:
Thanks for the edit @artarmstrong

PR #1662, opened 2017-10-19
Update index.md

anzcarroll:
description of the different types of joins


fvcproductions:
@anzcarroll Thanks so much for contributing changes on the `SQL` page! 🎉

For future reference, I would recommend focusing on formatting the content so it's easy to read and creating headers for new sections of content when it makes sense. 😊

PR #1668, opened 2017-10-19
Update index.md

hwallis:
More information on the `<div>` tag and contrasting it with a `<span>` tag.


Bouncey:
Thanks for the edit @hwallis

PR #1673, opened 2017-10-19
Description to why python

ashishsahu2008:


davcri:
Ok @ashishsahu2008 !  
Are you sure that the text can be used inside this project ? I found your article content here: https://www.quora.com/Python-programming-language-What-should-I-do-in-Python?share=1

Here it is terms of service of Quora: https://www.quora.com/about/tos

PR #1680, opened 2017-10-19
 add javascript-map-object

makkBit:
The map standard object of javascript and all its prototype methods are well explained.


Bouncey:
Thanks @makkBit, this is a great addition

PR #1686, opened 2017-10-19
Update control-flow

ctam91:
Add missing break and default statement.


Bouncey:
Thanks for the edit @ctam91

PR #1687, opened 2017-10-19
Defined open data, listed arguments in favor

srsullivan01:


Bouncey:
@srsullivan01 It seems that this article has had some changes already. Please review the current article and see if you can expand it any more. If you can, please raise a new Pull Request

Many thanks

PR #1688, opened 2017-10-19
First draft of Spike, with links

physicsjoe:


Bouncey:
Thanks for the article @physicsjoe 👍

PR #1689, opened 2017-10-19
added visual mode subtypes

vol-pi:


Bouncey:
Thanks for the edit @vol-pi 👍

PR #1691, opened 2017-10-19
Breadplop/branch1

breadplop:
slight edit from the previous branch


Bouncey:
thanks for the article @breadplop 👍

PR #1693, opened 2017-10-19
Link to using display grid from CSS Tricks

sbodi10:
Added JS fetch api link to understand how to make AJAX calls


Bouncey:
@sbodi10 We would like an article to describe and explain CSS Grid, not just display a link to an article/tutorial.

PR #1700, opened 2017-10-19
Added extra info on documentation

WillMcIntosh:
Added Arch Wiki to documentation as well as info on forums Ask Ubuntu and r/linuxquestions. Formatted that section for added clarity.


Bouncey:
Thanks for the edit @WillMcIntosh 

Sorry it took so long

PR #1706, opened 2017-10-19
Fix typos and add list of types

katyasyc:


Bouncey:
Thanks for the edits @katyasyc 👍

PR #1710, opened 2017-10-19
Add JavaScript/Arrow Functions article

daveaspinall:
A basic article covering the new ES6 arrow function expression syntax.


Bouncey:
Thanks for the article @daveaspinall 👍

PR #1714, opened 2017-10-19
Added bullet point

stolarz88:
Sketch is a good tool for designers that work with web developers often


Bouncey:
Thanks for the edit @stolarz88

PR #1721, opened 2017-10-19
Input

rutuke:


fvcproductions:
@rutuke, thanks so much for contributing changes on this page! 🎉

---

> For future reference, I would recommend using [Markdown links](http://google.com) instead of naked links like http://google.com. 🌐
>
> Also, it's good practice on GitHub to write a brief description of your changes when creating a PR. 📝

PR #1738, opened 2017-10-19
Create SSD.md

gwisk:
Stub for article on SSDs


erictleung:
If you're adding content, please format the article similar to more complete articles and add some relevant content.

PR #1753, opened 2017-10-19
Fix README.md

jpdilan:


texas2010:
thank you for the offer to fix it but will not accept it. sorry

PR #1770, opened 2017-10-19
Add a stub for python list comprehensions

hayleycd:
I have found these to be really useful, but as a fresh engineer had not seen them when I first started my job.

### DIFF
--- a/src/pages/python/python-list-comprehensions/index.md
+++ b/src/pages/python/python-list-comprehensions/index.md
@@ -3,11 +3,42 @@ title: Python List Comprehensions
 ---
 ## List Comprehensions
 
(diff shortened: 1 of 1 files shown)


texas2010:
@hayleycd do you plan to write about that?  or not. I will close this. because write article is better more than nothing of the stub.

hayleycd:
@texas2010 I would like to. I thought the procedure was to do a stub first (to get the topic approved) and then write the article. But if the topic looks good to you, I can write it as part of this pull request.

texas2010:
you don't need to get approved for that.   if you want to write about that. please do!

QuincyLarson:
@hayleycd Awesome - thanks for your contribution. Congrats on your first accepted pull request!

hayleycd:
@QuincyLarson thanks!

PR #1778, opened 2017-10-19
Format Specifiers

aidnurs:
Format Specifiers for printf and scanf functions


jp-sauve:

PR #1817, opened 2017-10-19
"improve" to "reduce

Afreen04:
Bug density should be reduced


fvcproductions:
@jp-sauve Thanks for correcting the wording. 

Going to go ahead and merge but @Afreen04 please consider adding more content in for future PRs or they will be marked as invalid. 😓

PR #1843, opened 2017-10-19
Included references to client and server side JS

MattMakes:
The article was lacking some defense for the language's true versatility. Also there were some corrections I've made to the prior entry regarding the term Versatility that was used instead of Interoperability, which I think the writer meant from the description they gave. Also updated article to call out when something was specifically a problem or gain if it was only seen in client browsers.


jp-sauve:

PR #1849, opened 2017-10-19
Fix link in "Free Code Camp Brownie Points"

jamesjpk123:


texas2010:
cant accept this PR. markdown link will be covert to html link in the auto script. so we dont need to change it. thank

jamesjpk123:
@texas2010 If you look at the article https://guide.freecodecamp.org/meta/free-code-camp-brownie-points, it seems as if the auto script is broken. It shows code in the article.

texas2010:
oh okay. make sense something affect on that. but the auto script is not broken.  so we could do it again.

PR #1870, opened 2017-10-19
Fix title of folder.

jamesjpk123:
Fixing title of the folder called "why-does-free-code-camp-uses-javascript-instead-of-ruby-or-python" to "why-does-free-code-camp-use-javascript-instead-of-ruby-or-python" to fix grammar, and be consistent with the title inside of index.md.


texas2010:
change a what? please put it in description. not PR title.

jamesjpk123:
Yeah sorry, here changing the title now.

texas2010:
it seems it is the same thing to me. can you tell me? I cant see what difference is.

jamesjpk123:
Changes the word "uses" to "use" to fix a grammar error. It was supposed to be fixed in #419, but @davidluhr forgot to change the folder name.

texas2010:
@jamesjpk123 thanks for explaining!

PR #1897, opened 2017-10-20
Update Mean, median, mode, and range

jiaw22:


ChaituVR:
Thanks for the Contribution. :+1:

PR #1914, opened 2017-10-20
Created "azure" folder inside cloud dev category

escottalexander:
we need the folder to begin filling it with information


texas2010:
you need to do that on local. make folder then index.md about that.

texas2010:
I close this since you make new PR....

PR #1920, opened 2017-10-20
Populate steganography security page

anasant:
I believe the path to the page is misspelt (stenography - which is shorthand writing). I'm not sure if y'all intended for the page to be about stenography, but steganography (sounds a lot like stegosaurus!) is a field in data security that I think you intended to include. 

Will try to update this path in a later PR.


johnkennedy9147:
@anasant 
 LGTM, thanks for your contribution.

PR #1924, opened 2017-10-20
Update index.md

MDodd423:
My previous change didn't seem to be the proper fix.  I believe this is the correct way to make the change.


texas2010:
yes that is correct.  good job!

PR #1929, opened 2017-10-20
Add Information on the Location Object

AJStacy:
Wrote an initial guide to the Location Object in JavaScript describing its use on elements and on the window and document objects.


jp-sauve:
Also, thanks for the accurate title and description.

PR #1941, opened 2017-10-20
Added changes

extrasad:


johnkennedy9147:
@extrasad Hi, I'm closing this PR as the edits are too minor and per the contribution guidelines links should be in markdown form rather than html anyway so they will all get changed eventually

PR #1958, opened 2017-10-20
Added vector iteration to guide

romogo17:
Added vector iteration to guide. In the 3 most common ways: the first one involves using a range-based expression (since C++11), the second one uses iterators, and the last one is a index-based iteration


jp-sauve:
Hey @username, thank you for your contribution to the page! :+1:

PR #1962, opened 2017-10-20
New article: Passing pointers to funtions

RajatGarga:


jp-sauve:
This seems like it would be a good candidate for inclusion in https://guide.freecodecamp.org/c/pointers instead of being its own article.

QuincyLarson:
@jp-sauve Pointers and their use are pretty complicated, and wouldn't fit into a single article. I say we err on the side of specificity and more articles, not fewer. This looks great, so I'm going to go ahead and merge it.

jp-sauve:
Hey @RajatGarga, thank you for your contribution of the new page! :+1:

PR #1976, opened 2017-10-20
Close a missing parenthesis

german1608:


johnkennedy9147:
@german1608 minor but important!

PR #1985, opened 2017-10-20
Updated README.md with more dev info

hisivasankar:
Summary:
Dev guide to set up the site locally was missing some key information about what happens after you run `yarn run dev`. Hence updated the same so that it will be useful for beginner developers.


IsaacAbrahamson:
Changes LGTM. Thanks for contributing!

PR #1999, opened 2017-10-20
Relate content to CSS properties

sethlu:
Removed extra information about background image since its painting area can be defined with `background-clipping` and doesn't have to be only inside the padding area.


jp-sauve:
Hey @, thank you for your contribution to the page! :+1:

PR #2018, opened 2017-10-20
Added depth to initial version

terrenjpeterson:
Expanded on simple definition of the term, including providing alternative model (aka throw it over the wall).


Asjas:

PR #2032, opened 2017-10-20
UPDATE query with joins

iamnemesis:
Added a guide to write an update query with joins


Asjas:

PR #2060, opened 2017-10-20
add links

Juell1635:
[I'm an inline-style link](https://www.google.com)

[I'm an inline-style link with title](https://www.google.com "Google's Homepage")

[I'm a reference-style link][Arbitrary case-insensitive reference text]

[I'm a relative reference to a repository file](../blob/master/LICENSE)

[You can use numbers for reference-style link definitions][1]

Or leave it empty and use the [link text itself].

URLs and URLs in angle brackets will automatically get turned into links. 
http://www.example.com or <http://www.example.com> and sometimes 
example.com (but not on Github, for example).

Some text to show that the reference links can follow later.

[arbitrary case-insensitive reference text]: https://www.mozilla.org
[1]: http://slashdot.org
[link text itself]: http://www.reddit.com

### DIFF
--- a/README.md
+++ b/README.md
@@ -210,6 +210,30 @@ With your help, we can create a comprehensive reference tool that will help mill
 
 Copyright (c) 2017 freeCodeCamp.
 
+
+
+[I'm an inline-style link](https://www.google.com)
+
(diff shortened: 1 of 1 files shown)


texas2010:

PR #2113, opened 2017-10-20
Update Units article.

harryray:
Now contains info on `px`, `vw` and `vh`.


Asjas:
We're happy to accept these changes, and look forward to future contributions

PR #2220, opened 2017-10-20
add new article for git bisect

samuel-hu:
New page for the `git bisect` command.


HKuz:
This LGTM! ⬅️ 2️⃣ ➡️  I wasn't aware of the Git bisect functionality, thanks for adding a great article about it, @samuel-hu !

PR #2221, opened 2017-10-20
Useless quotes in the object syntax

QuentinChevance:


jp-sauve:
I'm closing this for now, as the change neither corrects an actual fault, nor adds content that was not there. If you would like to amend the PR with content that clears up the differences between quotes and no quotes, you can reopen it, and we will re-evaluate at that point.

PR #2230, opened 2017-10-20
Update index.md

twilliams1615:
Added button element and datalist option


twilliams1615:
adding the datalist and button elements

PR #2237, opened 2017-10-20
Add if statement on C

luigicapogrosso:


src/pages/c/if/index.md
```diff
@@ -0,0 +1,64 @@
...
+```
+
+When `condition` is true, code inside the `if` section executes, otherwise `else` executes. Sometimes you would need to add a second condition. For readability, you should use a `else if` rather than nesting `if` statements.
+
+```
+if (condition) {
+	// Do something if `condition` is true
+}
+else if (anotherCondition) {
+	// Do something if `anotherCondition` is ture
```
Oxyrus: Do something if `anotherCondition` is **true**

src/pages/c/if/index.md
```diff
@@ -0,0 +1,64 @@
...
+}
+else {
+	// Do something if `condition` AND `anotherCondition` is false
```
Oxyrus: // Do something if `condition` AND `anotherCondition` are false

Asjas:
LGTM 👍. Merging PR. The spelling mistakes and usage of nouns can be fixed in a future PR. @luigicapogrosso Thanks for the PR. :memo:

PR #2297, opened 2017-10-20
Create NQueen.cpp

mintoo511:


texas2010:
@mintoo511 that is enough. stop try hurt guide repo.

PR #2303, opened 2017-10-20
Updated More Information section

ampc:
Fixed broken npm wikipedia link


texas2010:
thank but someone did fix it.

PR #2312, opened 2017-10-20
Ethical Breaches: Fixed link formatting - fixes #818

arcdev1:
Eliminated erroneous space between link title (square brackets) and url (parens) so that links display correctly.
Also updated "More Information" so that it is an unordered list (bullets).

cc: @davcri


davcri:
LGTM! Thank you for the pull request!

PR #2314, opened 2017-10-20
update matrices intro

lhcavalcanti:
What is a matrix, and the basic operation (+, -, transposition).


13thThief:
~~Above~~ Below

PR #2316, opened 2017-10-20
Added for loop alternative

rjanecek:
Acceptable alternative is to declare the variable within the for loop instead of using two different statements.


Asjas:
We're happy to accept these changes, and look forward to future contributions. :memo:

PR #2324, opened 2017-10-20
loops.md

saboorElahi:
It introduces the for loops in C++


johnkennedy9147:
@QuincyLarson @Bouncey @systimotic Shouldn't this have failed build tests? -src/pages/cplusplus/loops.md

systimotic:
@johnkennedy9147 This doesn't break the build, as the `cplusplus` directory already has an `index.md`, so I guess the tests were technically correct 😄 You're right though, any file that isn't named `index.md` is ignored, so this file wouldn't show up in the navigation. We might want to update the tests to fail on files with other names.

saboorElahi:
Then how to make a folder in some directory. Like there is no directory named loops in cplusplus directory. How do we upload a for loop file in this directory ? And what name should we give to our file ?

saboorElahi:
Ok. Thanks. I have also pushed some files whose name is not correct in the past as I don't know the problem.
Will try to name the file correctly in the future!

johnkennedy9147:
@saboorElahi Hi could you fix the path and add the front matter to your 3 open PRs.

First 3 lines must contain front matter in this format:
```
---
title: title goes here
---
```

thanks

saboorElahi:
Yes sure

PR #2352, opened 2017-10-20
Update index.md

codeforgeek:


Samuel-Clarke123:
Note: I fixed a bug whilst I came through, as you should be looping from` i=1` to the end, not `i=0`, as you are using `[i - 1]` later on. `( (i = 0) (- 1 = -1)`, which is not an index).

PR #2375, opened 2017-10-20
Added space complexity to quick sort article

nmbrgts:
Space complexity is a big selling point for quick sort. So, I thought it would be a good addition to the article. 

Also, I changed a small type in the example code.


QuincyLarson:
This looks solid. Thank you for your contribution!

PR #2393, opened 2017-10-20
Content for Ansible

nikodonnell:
Here is some content for Ansible that I've just written, we actually use Ansible in our stack, so I've 'used' it without fully understanding it - so it was nice to dig deeper into it.


jp-sauve:
Hey @nikodonnell, thank you for your contribution to the page! :+1:

PR #2409, opened 2017-10-20
fix resource link formatting

JQuinnie:
formatted resource links into an unordered list and structured links for readability.


texas2010:
cant accept this PR because markdown link will be covert to html link in the auto script. so it is pointless to change that again. sorry!

PR #2425, opened 2017-10-20
Wrote the Iframes article

nai888:
I wrote the article, including multiple code examples. I also put some HTML directly into the document to display those examples, but I am unsure if HTML renders correctly in .md files. If not, the examples will need to be removed.


nai888:
@nsuchy No worries; looks like maybe lots of people submitted pull requests for this article before any of them had been merged in.

PR #2427, opened 2017-10-20
Patch 1

StevePaytosh:
Removed the misspelled stenography file and replaced it with a steganography file.  The new index file is expanded from the original.


johnkennedy9147:
@StevePaytosh  Hi, sorry but there was already a pending PR for this article.  Please review the article and propose any additions or edits you think would improve it. thanks for your contribution.

PR #2446, opened 2017-10-20
Wrap opening paragraph `<br>` in back-ticks

erunyon:
Without the back-ticks, the `<br>` was functioning as an actual break instead of displaying the html.


johnkennedy9147:
@erunyon Good catch! Thanks for the contribution.

PR #2455, opened 2017-10-21
Clean up formatting

erunyon:
- Remove unnecessary semicolons
- Make more info a list


texas2010:
change link back.  btw markdown link will be covert to html link in the auto script.

PR #2458, opened 2017-10-21
Update README.md

parikhparth23:


texas2010:
thank you for offer to fix it. but not accept this time.

PR #2460, opened 2017-10-21
Formatting in the guide was rendering as html

erunyon:
The guide page was displaying as html instead of displaying the code. Switched the opening back-ticks to include 'html'


Hijerboa:
LGTM! Thanks for your contribution!

PR #2463, opened 2017-10-21
Added the border radius property and defined it

Maluchukwu:


davcri:
@Maluchukwu thank you for your pull request !

PR #2473, opened 2017-10-21
Added code example using the tag.

ezebecke:


texas2010:
@ezebecke can you tell me why there have other same examples?  it seems it is the same thing to me.

ezebecke:
@texas2010 yes! Because one it would be how the HTML code works and the other one how it would look.

1andee:
I agree with @texas2010. The example in this PR is virtually identical to the existing one.

PR #2511, opened 2017-10-21
Add C syntax highlight to the pointer guide

andres-uris:


Ethan-Arrowood:
This LGTM. We should be verifying that posts coming in have the language on preformatted blocks defined, but sometimes things are missed.

PR #2534, opened 2017-10-21
Change output from 2 to 2.0

BenBurbank:
Output from the double array was set at 2 and should be set at 2.0 (double).


Hijerboa:

PR #2593, opened 2017-10-21
Async messaging with Node.js & RabbitMQ.

rudimk:
Added an article on async messaging with Node.js and RabbitMQ using Tortoise.


Hijerboa:

PR #2613, opened 2017-10-21
Enhanced operating steps and purged planning

quiteBold:
It is a moderation technique for estimating only. Not for planning. 

Added step of PO explaining the feature to the team, which is essential. 
Added step of lowest and highest estimators discuss, which leads to better understanding and refinement.


Hijerboa:
LGTM! Thank you for correcting that!

PR #2614, opened 2017-10-21
Html doctype declaration

GonzaloCBenlloch:


Hijerboa:
@GonzaloCBenlloch unfortunately, it looks like someone else has already written this article. Sorry! We hope you will be able to help us right others in the future. :)

PR #2624, opened 2017-10-21
Added let declaration.

beawolf:


jp-sauve:
The "let" declaration is ES6 syntax, and is not specific to typescript. I'm not a typescript user though, so if you know of a compelling reason for this to be merged here, please reopen the PR, and let someone know in the Gitter Contributor's chat that you'd like someone to take a look. The page I believe this content belongs is https://guide.freecodecamp.org/javascript/es6/let_and_const  if you would like to work your content in to the existing content on that page. Thanks for your hard work, and I hope to see more contributions from you in the future!

PR #2628, opened 2017-10-21
Added const declaration.

beawolf:


jp-sauve:
Thank you for your pull request. This content belongs elsewhere, so I am closing. Your effort is appreciated, and if you'd like to work your changes in to the existing content in the ES6 section, we can then try and merge them in. Same as #2624

PR #2640, opened 2017-10-21
Added information on distributions

illninofan:
Added information on distributions and listed a couple, along with package managers and some info on using Linux distros via Live CD or USB


Hijerboa:

PR #2648, opened 2017-10-21
expanded the article stab

akolybelnikov:
the concept does not require much explanation besides the definition and a few short examples


Hijerboa:

PR #2653, opened 2017-10-21
Removed unnecessary '[' in JS comment

odv:
'[' has no purpose in this JS comment and could confuse some of the beginners reading the guide.


Hijerboa:
LGTM! Thank you for finding and fixing this!

PR #2666, opened 2017-10-21
Removed debugging

jlsajfj:
Out of place, deserves its own page


jp-sauve:
Really, this whole page deserves a rewrite. From the unfortunate naming, to the lack of focus, it could use a lot of work. I'm closing this PR because identifying this part of the content as misplaced, and then removing instead of moving it, doesn't really improve the guide sufficiently.

PR #2674, opened 2017-10-21
ES6 rest parameter

MrNime:
Some more info about the use of the ES6 rest parameter as alternative to using the arguments object


jp-sauve:
Hey @MrNime, thank you for your contribution to the page! :+1:

PR #2675, opened 2017-10-21
Replaced W3schools resource with MDN

odv:
First of all, 'W3C - HTML DOM' is a link title that could confuse beginners to the point where they will think that W3schools website is affiliated with https://www.w3.org/. As it says on https://www.w3schools.com/about/default.asp they have no affiliation with W3.

Also, please see: http://www.w3fools.com/

I've replaced that link with a MDN resource and think that we should stop referencing to that site on all guide articles.

### DIFF
--- a/src/pages/javascript/html-dom/index.md
+++ b/src/pages/javascript/html-dom/index.md
@@ -1,7 +1,7 @@
 ---
-title: HTML Dom
+title: HTML DOM
 ---
-## HTML Dom
+## HTML DOM
 
 With the HTML DOM, JavaScript can access and change all the elements of an HTML document.
 
@@ -23,5 +23,5 @@ With the object model, JavaScript gets all the power it needs to create dynamic 
 
 #### More Information:
 
-[W3C - HTML DOM](https://www.w3schools.com/js/js_htmldom.asp)
+[DOM](https://developer.mozilla.org/en-US/docs/Web/API/Document_Object_Model)
 


src/pages/javascript/html-dom/index.md
```diff
@@ -23,5 +23,5 @@ With the object model, JavaScript gets all the power it needs to create dynamic
 
 #### More Information:
 
-[W3C - HTML DOM](https://www.w3schools.com/js/js_htmldom.asp)
+[DOM](https://developer.mozilla.org/en-US/docs/Web/API/Document_Object_Model)
```
jp-sauve: I agree that MDN is a better resource than W3Schools, but why should a beginner resource not include references to the latter? Even the w3fools page suggests that MDN should always be included, but not that W3schools should be ignored. If you feel you have a good case, open an issue, and reference it here. Or, if you bring the original link back (fixing it's anchor text, as I agree that it isn't correct as is), I think we could merge.

odv: @jp-sauve  Personally, I think it's really important for beginners to learn how to read documentation and be professional. 
I think I've made it quite clear in the message above. W3schools could often trick people into thinking they are affiliated with W3 and in my opinion was only built for monetization/certificate selling.
Since you have write access to this repo, it's your call.

johnkennedy9147: I disagree with the idea that w3schools is not an appropriate site to link to - it provides good and easy to understand tutorials on a load of topics. More technical sites can be daunting for beginners. This decision should probably be referred to @QuincyLarson for a decision. It would be a significant task to remove all links already in guide and preventing future contributors from linking to it would an additional burden for reviewers.

QuincyLarson:
@odv The MDN is definitely better than w3schools, but w3schools has slowly improved over the past few years. I don't think it's a problem to link to it. Thanks for helping improve these guide articles, though. We welcome future contributions from you.

PR #2682, opened 2017-10-21
Added Key-value databases article

oribaldi:
A short introduction for key-value databases


adiparab14:
@oribaldi Thanks for your contribution 👍

PR #2683, opened 2017-10-21
Replaced typographic quotes

quiteBold:
There were some typographic quotes like ”“ instead of plain double quotes " which could cause trouble for people copying.


jp-sauve:
Hey @quiteBold, thank you for your contribution to the page! :+1:

PR #2709, opened 2017-10-21
Update index.md

rainyramteke:


davcri:
Hi @rainyramteke, unfortunately we cannot accept your contribute because it is mainly copy pasted from other sources and this goes against our [Academic Honesty policy](https://www.freecodecamp.org/academic-honesty).

You are allowed to cite other sources, giving appropriate [attributions](https://github.com/freeCodeCamp/guides/blob/master/CONTRIBUTING.md#attributions). 

Let me know if you have any question.

PR #2762, opened 2017-10-21
Create installing-go

santos22:
Hi!

I wrote this guide because I know I struggled mightily when first installing Go on Windows, setting GOPATH, adding Environment Variables, etc. I wrote this guide in hopes of making it easy for individuals to continue their journey to writing Go and not being turned away by something like installation that should be simple 😄


Hijerboa:
Great guide! Thank you for contributing this--we hope to see you again! :)

PR #2766, opened 2017-10-21
Changed some parts to match Article Style Guide

vromaniv:
- replaced passive voice with active voice;
- now text uses the second person ("you").


Hijerboa:
LGTM! Thank you for fixing this!

PR #2825, opened 2017-10-21
added benefits of switch

davidemily:
added benefits of using a switch statement vs if/else and a line saying that defaults do not need a break statement.


Hijerboa:

PR #2834, opened 2017-10-21
Created a file on malloc

davidemily:
Malloc is an important part of the C language so I thought it needed to be discussed.


Hijerboa:
Awesome article! Thank you for writing this! :)

PR #2839, opened 2017-10-21
Added HOF and map as an example

OneCodeMan:


Hijerboa:

PR #2856, opened 2017-10-21
Added Integration-Rules Article to Replace Stub

Lest-V:
I created an article with the most common integration rules to replace the existing stub. These are only the basic/common rules, but the linked pages have more complete lists.  The additional rules are harder to type cleanly due to complex fractions. 

I chose to use additional spaces between portions of the formulas so that the different elements are clearer.


Hijerboa:
LGTM!  Thank you for contributing!

PR #2866, opened 2017-10-22
Fix github Markdown formatting bug

DannyAllegrezza:
Currently, the C# page simply renders "C" as the title. This is due to how markdown handles the "#" symbol. This PR will fix the title so that it properly shows up as `C#`

ref: https://stackoverflow.com/questions/32196555/how-to-escape-the-hash-sign-in-a-github-markdown-header-backslash-is-not-w

### DIFF
--- a/src/pages/csharp/index.md
+++ b/src/pages/csharp/index.md
@@ -1,7 +1,7 @@
 ---
 title: C#
 ---
-## C#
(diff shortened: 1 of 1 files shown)


adiparab14:
Thanks for your contribution :+1:

PR #2884, opened 2017-10-22
Dunning Kruger Effect

jjpatel361:
Don't suffer from **`Dunning Kruger Effect`**!


Hijerboa:
Great article! Thank you for writing this! :)

PR #2886, opened 2017-10-22
Fixed minor typo

Kanen31:
"upto" changed to "up to" which, from my understanding, should always be two words even when used with numbers/percentages.


Hijerboa:
LGTM! Thank you for catching this! :)

PR #2889, opened 2017-10-22
Add BSD OSes entry (like Linux entry)

tekbasse:
Adds Application Frameworks, BSD OS, real-time OS, Robotics entries


tekbasse:
I'm temporarily closing this. Will resubmit. This PR accidentally includes incomplete info that has already been revised and approved separately.

PR #2898, opened 2017-10-22
Added guide for python list map reduce and filter

hafiz703:


Hijerboa:

PR #2899, opened 2017-10-22
Minor grammatical changes

thisaaronm:
Made small grammatical changes, sometimes for consistency purposes. Emphasized "2" and "3" in the Arithmetic Operators section because they are being referred to as components of the example expression (which I think probably should be backticked too, but I left that alone), similar to the +.


Hijerboa:
LGTM! Thank you for fixing these!

PR #2903, opened 2017-10-22
cooling stub

freedog96150:
Added a cooling stub with some starter information


freedog96150:
Delete failing PR to fix and resend.

PR #2917, opened 2017-10-22
Added a folder for jquery event method

Neeeophytee:
These methods are used to register behaviors to take effect when the user interacts with the browser, and to further manipulate those registered behaviors.


Hijerboa:
Hey @Neeeophytee would you please write an article to go along with this?

PR #2936, opened 2017-10-22
Fixing a line on this page

dreamer-nitj:
Fixed the first sentence on this page to make it more readable.


raisedadead:
@dreamer-nitj thanks for the pull request. 

I think the verbiage is correct in stating that 

> `git pull` is a Git command ...

because, it should be assumed that anyone can land on this page from anywhere and introducing this as a Git command is stating that notion explicitly.

I'll leave this open for others, to take a look. Feel free to comment and improve.

Happy contributing!

PR #2972, opened 2017-10-22
Added tools to create SVG

Neeeophytee:
When I started out as a developer, I needed a graphic tool to create SVGs, so the index file should have a reference to all tools available. Added description to Inkscape and Adobe Illustrator as they are the most used ones.


jp-sauve:
@Neeeophytee 

Issue #310, opened 2017-09-23
Translation to other Languages 


ahasna:
Are you guys considering to translate your content to other languages? I am happy to start doing this to my language Arabic

erictleung:
@ahasna thanks for the issue! This could be great. Let's hear from others on how we could or should implement translations.

cc/ @QuincyLarson @Bouncey

JuanLuisCh:
I'm able to start translating into Spanish, but, first we need to solve one thing. How to add the articles in different languages without being redundant

evaristoc:
@JuanLuisCh @ahasna:

The following below are my personal opinions:

It is important to take in account what @systimotic mentioned in the Contributors channel: there is currently no plans to make the translations available as the first objective would be to settle the existing project. However IMO translating while the material in English is produced would be a big advantage and will easy the process of versions in different languages when the technical challenges get solved.

So I am, like @erictleung and many others, absolutely positive about your initiative.

If your initiative project receives a positive feedback I would suggest the following:
1. Start getting organized. You might end up been the leaders of possible translation efforts and focal points if more people want to get involved or need information. Prepare to organize the help and a translation plan, so look for info about how to manage a small team.
2. Establish the targets. In my opinion, it would make more sense if the translation team focus on the technically easiest parts first? Focus on an audience that would benefit for start learning the basics in your countries. Example: SCHOOLS (children), computer-illiterate audience or organic searches (like looking for wikipedia article in your native language).
3. Think on the translations as an extension not only for FCC course, but for ANY existing online and OFFLINE course, so it would be great if the selection of the translated material reflect the needs of your target country.
4. I would invite you to suggest material in English that could benefit your audiences.

If you start translating EVERY material you can:
1. Waste a lot of effort on material that might not reach your target audience, making it less effective
2. It will be harder for you and any team built around the translation effort.

@JuanLuisCh it could be redundant if your target audience has already English as language. The effort should be made on revealing information to an audience for which English is still not enough, that benefit for partial or total translation to understand the topics, and that could benefit from start learning web development. I can confirm that that target group exists.

evaristoc:
@ahasna @JuanLuisCh 
Don't get me wrong regarding the point about "you might end up been leaders...". It is not that you will lead it. You can either lead or be part of a group of people working on it. So don't get scared. The fact is that you are the first I know who are proposing translations - there is no translation team at the moment.

[...]

PR #2985, opened 2017-10-22
Make index for indonesia language

karindralinux:


fvcproductions:
Hey again @karindralinux 

I will get back to you on this because as of right now, there is no support for other languages. 

This is a feature that will probably be worked on in the future but thank you again for your contribution! 😊

fvcproductions:
The ability to translate this content into other languages is currently on the roadmap - see https://github.com/freeCodeCamp/guides/issues/310. 

For the time being, I will close your PR but mark it with a `translation` label so that it may be re-opened in the future after this feature is actually implemented.

PR #2990, opened 2017-10-22
a method for converting decimal into fraction is documented

9633089281:


jp-sauve:
@9633089281 

PR #104, opened 2017-07-29
Added canvas pages

alan-luo:
Added a bunch of canvas pages. Please review and let me know if any changes are needed.

PR #3007, opened 2017-10-22
Add canvas (take 2)

alan-luo:
Adding canvas pages.  Initially started in #104. Sorry for the huge delay on this - I was way busier during the fall than expected, but my work load is finally starting to come off a bit. I would like to add more to these pages at a future date, but for now I just don't want to leave these hanging.

Other thing: it would seem that the articles have been restructured since I last touched this. I've moved the articles out into the parent folder - let me know if this was not the correct move.

Sorry for the delay ~ happy to be back. 😄


Asjas:
@alan-luo 

Thank you for your contribution:+1:. This is a really in-depth and extensive addition you've made 💯 

PR #3034, opened 2017-10-22
update 'black box testing'

georgiebin:
updated black box testing with additional info. also added an introductory site on testing to the resources


ChaituVR:
Thanks for the PR!

PR #3040, opened 2017-10-22
Query parameterization (benefits, sample code)

grantwinney:
I described the basics of query parameterization and the potential pitfall of not using it, including code samples.


Asjas:

PR #3082, opened 2017-10-22
Added an example

codaitya:
Added an example to demonstrate absolute value on negative numbers exponentiatiob


ChaituVR:
Thanks for the PR! :+1:

PR #3090, opened 2017-10-22
added extra explanations

JoeSuzuki:


Asjas:
@JoeSuzuki 

Thank you for your pull request but I don't feel that this is a substantial contribution. 

Changing from the original number 10 to a undeclared variable n would cause the loop to fail. Adding the comment is also unnecessary. 

I will be closing this PR. Please let me know if you have any questions.

PR #3172, opened 2017-10-22
Update index.md

1minPeru:
Added String data type, explained how it is not primitive and used manipulation of the data type in an example


Asjas:

PR #3180, opened 2017-10-22
PHP Classes and Objects

tonydelanuez:
Explained the basics of PHP classes and objects, along with private variables, instance variables, methods, etc.


ChaituVR:
Thanks for the PR! :+1:

PR #3198, opened 2017-10-23
real world example

ryanator5000:
This gives an idea that is easier to associate expansion cards with modern application.


ChaituVR:
Thanks for the PR! :+1:

PR #3206, opened 2017-10-23
fixed console.log on "Common Pitfalls" example

fernandovbs:
The **Exceeding the bounds of an array** example have to print the array position value based on the index.


ChaituVR:
Makes sense, Thanks for the PR! :+1:

PR #3220, opened 2017-10-23
Added Parallel Computing topic

jennifertran:
Parallel computing is a type of computation in which many calculations or the execution of processes are carried out simultaneously. This topic can be applied to computer science.


Asjas:

PR #3249, opened 2017-10-23
Add link to WAVE for viewing pages without CSS

ryan-p-randall:
Using the WAVE link might be easier/faster for people still learning browser developer tools.


ChaituVR:
Thanks for the PR! :+1:

PR #3283, opened 2017-10-23
Add .each method explanation and example

ryandav:


ChaituVR:
Thanks for the PR! :+1:

PR #3287, opened 2017-10-23
Simplified Hello World

sprajosh:
Things like `if __name__ == "__main__"` can be taught at later stages because many students try Python as their first programming language and this statement makes it look much more complex than it actually is.


ChaituVR:
Thanks for the PR! :+1:

PR #3295, opened 2017-10-23
Inline to reference style images

Maartz:
Changing images inline-style to reference-style


Asjas:
Thank you for your pull request but I don't feel that this change is needed. 
I will be closing this PR. Please let me know if you have any questions.

PR #3298, opened 2017-10-23
Added One common mistake

AmarNathH:
Added a common mistake for accessing the last element of an array


ChaituVR:
Thanks for the PR! :+1:

PR #3337, opened 2017-10-23
take, drop methods added

rahul-das:
take, drop methods added for ruby array


Asjas:

PR #3338, opened 2017-10-23
Add info about typeof NaN

MrNime:
Added some info about the output of `typeof NaN`
`typeof NaN` results in `number`, which is counterintuitive.


Asjas:

PR #3374, opened 2017-10-23
Write a small article about PHP variables.

NathanHeffley:
I love working with PHP, so I figured I would start a series of PHP articles with this simple intro to how variables work in PHP.


Asjas:

PR #3393, opened 2017-10-23
pointer as a function parameter

Vatspra:
how to use pointer as a function parameter


Asjas:

PR #3398, opened 2017-10-23
add .join method to document

robert-laws:
add the .join method to the basic array methods


Asjas:

PR #3408, opened 2017-10-23
Updated "grooming" with "refining"

artismarti:
"Refining" is the new & recommended term because of other (not so innocent) connotations of the word "grooming". :)


Asjas:

PR #3435, opened 2017-10-23
Created a-simple-web-server-in-go

mpboom:
Well, my first pull request ever.

I wrote a bit about how to create a web server in Go. I am planning to write some more articles about Go, but I would like to see if I did everything right with this one. This way I hope I won't make the same mistake twice.

Feedback is always welcome and appreciated.

### DIFF
--- a/src/pages/go/a-simple-web-server-in-go/index.md
+++ b/src/pages/go/a-simple-web-server-in-go/index.md
@@ -1,0 +1,30 @@
+---
+title: A simple web server in Go
+---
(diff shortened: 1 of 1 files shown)


jp-sauve:
@mpboom 
For future reference, one improvement would have been using "#" to mark section headers. Your submission isn't so long that you would need many, but a main header at the top of the page would have been good. I'm merging unedited though, and we can clean it up later. Otherwise, nice work, and congratulations on a successful first commit!

mpboom:
Ah, I misunderstood a bit and thought that the
```
---
title: A simple web server in Go
---
```
would automatically generate a first heading!

Of course, still thank you very much for merging!

PR #3495, opened 2017-10-23
Added content to Bootstrap Forms article

leticiabecker:


leticiabecker:
Hi Ian!

You are absolutely right! I'm sorry for that!

How could I fix it, please?

Thank you!
ᐧ


-----
Leticia Becker

1andee:
Howdy @leticiabecker 

Because there was a boatload of Pull Requests in October (and over 4,000 open ones in the queue), there's been a ton of changes since you originally forked your repository. So, first, you'll want to sync your fork if you haven't already done so.

Using the command line, you can do this in three easy steps:

```bash
git remote add upstream git://github.com/freeCodeCamp/guides.git

git fetch upstream

git pull upstream master
```

If you're using a GUI, you can simply `Add a new remote...` and use the link `git://github.com/freeCodeCamp/guides.git` from above.

Once the above step is complete, you could either:
1) Start a new branch and make a new Pull Request with edits to the correct article (this is probably the easiest solution that I would personally endorse); or
2) Using the same branch (I believe it's `branch-3`), make a new commit that does the opposite of what you've proposed here, followed by further commits with edits to the correct page.

Feel free to ask if you have any questions!

PR #3503, opened 2017-10-23
Corrects the width and height CSS for the Images

YoKoGFX:
Removed the PX as it is not correct.

http://w3c.github.io/html/semantics-embedded-content.html#the-img-element


thomaskuntzz:
Hi @YoKoGFX. Congrats on your first pull request (PR)! :tada:

Thanks for fixing this!

PR #3526, opened 2017-10-23
Update Python Lambda expressions article

p-m-m-c:
Included description and three examples


thomaskuntzz:
There's also #8027 for this aritcle

PR #3527, opened 2017-10-23
Fix explanation for JS logical operators & add examples

arnav-aggarwal:


thomaskuntzz:
Hi @arnav-aggarwal. Thanks for your contribution!

I thought it would be good to keep the more simple explanation, so while resolving merge conflicts I tried to find a compromise by leaving both the more technically correct explanation you added and the more simple explanation for when dealing only with booleans.

Could you tell me what you think? 🙂

arnav-aggarwal:
I like it!

thomaskuntzz:
Cool, then let's merge this!

PR #3529, opened 2017-10-23
Updated content and gramer

andrewehipp:
Rewrote most content to be more accurate. Updated external example to look better.


thomaskuntzz:
Thanks for your contribution!

PR #3554, opened 2017-10-23
Update UTF 8 article

r14n:


Bouncey:
Thanks @r14n! sorry for the long delay 👍

PR #3574, opened 2017-10-23
"Earn certificates and boost your hireability" in certificates article

kbishop6:


thomaskuntzz:
Hi @kbishop6, thanks for your contribution, but this is not something we'd like to merge as it's not a substantial contribution.

PR #3672, opened 2017-10-23
Add HTML Font Color Attribute

armen-gabrielyan:
Added content and example about HTML font color attribute.


Palmidence:
I think saying `<font color>` as the attribute may be inaccurate, since it's not `</font color>`. Perhaps just say `color`

PR #3739, opened 2017-10-24
Add short info for HTML Entities

stmapman7:
Provide definition of HTML entities and additional info link from a helpful resource.


stmapman7:
HTML entities page has been updated.

PR #3763, opened 2017-10-24
Add a few stub pages for Redux

abustamam:
I plan to write articles for at least three of these topics (reducers, actions, and reselect). I've added a few stubs for topic ideas that I think people should understand when diving into the redux ecosystem:

- redux sagas
- redux thunk
- redux selectors (not using `reselect` library)
- redux middleware (conceptually)

Hopefully someone else will be able to tackle those four topics, though I may do so myself in the near future.

### DIFF
--- a/src/pages/redux/redux-actions/index.md
+++ b/src/pages/redux/redux-actions/index.md
@@ -1,0 +1,15 @@
+---
+title: Redux Actions
+---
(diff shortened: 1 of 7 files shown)


Bouncey:
Thanks @abustamam Look forward to these articles

PR #3784, opened 2017-10-24
fix some syntax errors in code examples

abustamam:
There was an issue with one of the code examples. It used `const balance = ...` and later, `balance += ...`. If run in a terminal, this would throw an error: "Reassignment to const variable."

I have opted to use ES5-compatible syntax over ES6 syntax for ease of use for those who do not know ES6. Additionally, I've named a few variables more appropriately to make it easier for students to visualize. 

Also fixes some minor grammatical errors.

### DIFF
--- a/src/pages/javascript/closures/index.md
+++ b/src/pages/javascript/closures/index.md
@@ -13,13 +13,13 @@ function by(propName) {
     }
 }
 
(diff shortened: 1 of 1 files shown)


Bouncey:
@abustamam Thank you for the PR, but I don't think these edits are necessary.

PR #3788, opened 2017-10-24
Update index.md

cheeserulz:
remove part about "maximo" since it's not  part of the dataset and just confusing / distracting


jamesjpk123:
@ericleung looks like CI checks are done!

PR #3836, opened 2017-10-24
Changed the "for loop" section

andrew0475:
Changed the definition of the third part of the for loop from "afterthought" to "increment/decrement"


Hijerboa:
LGTM! Thank you for contributing! :)

PR #3887, opened 2017-10-24
Add classes guide

petarblazevski:


domwakeling:
Hey @petarblazevski 

This PR doesn't actually explain what a class is, and a lot of the content appears to be taken from another source without attribution 😞

Suggest it could be re-written to match the style of the other *-type guides in the Typescript section

petarblazevski:
@domwakeling thanks for taking look at this. Looking at the other guides in the Typescript section, I don't really see a lot of explanation either. I agree this could be rewritten with a much better explanation. And I am willing to do that. But please consider removing the rest of the guides without a good explanation.

In the following days, I will update this PR with a better content

domwakeling:
@petarblazevski thanks; I don't think we remove content once accepted, but maybe you could look at improving them under a new PR?

petarblazevski:
@domwakelingI will look into into improving the other guides as well

PR #3891, opened 2017-10-24
Update article in Javascript/Promises

davidkwan95:
Removing the example library of redux-saga because we are not talking about Redux. The reason for this is not to add more confusion when readers are reading about Promise.


ChaituVR:
Hi @davidkwan95 . Congrats on your first pull request (PR)! 🎉
I think this is correct, it will be confusing when readers are reading about Promise

PR #3942, opened 2017-10-24
Expanded description, font family and font weight.

rblakejohnson:


farverio:
Looks great, thanks for contributing!

PR #3960, opened 2017-10-24
add my changes

aman-009:


farverio:
Hi @aman-009, thanks for contributing! This article looks pretty similar to https://www.cis.upenn.edu/~matuszek/cit594-2012/Pages/backtracking.html.

Do you mind re-writing so we can merge?

PR #4080, opened 2017-10-24
added my changes

prabusubra:
added the information about promise immutability


prabusubra:
added information about promise immutablity

PR #4099, opened 2017-10-24
Update index.md

AllenCompSci:
Not a change that is absolutely needed here, but one in the folder section.


domwakeling:
Hey @AllenCompSci I think this is more appropriate as an Issue rather than a Pull Request so am closing

Thanks for your contribution

Issue #897, opened 2017-10-15
Fixing some a11y, need React guidance


svinkle:
Hi there! I'm attempting to fix up some accessibility issues with the site itself and need some help with understanding this React code. Specifically, the elements which render the side nav panels.

I've narrowed it down to [`renderHeader()`](https://github.com/freeCodeCamp/guides/blob/master/src/LayoutComponents/nav/NavPanel.jsx#L76) where this issue seems to be. This method returns a `div` with some `span` elements, but on component render, the `span`s are children of an `a` element. These `a` elements are using the appropriate `aria-*` attributes to help convey state and context, but there are a couple issues:

1. The `a` elements are missing an `href` in order to receive keyboard focus
2. The `onClick` handler is on the parent `div` element. This needs to be set on the `a` in order for keyboard users to activate the panel.

I can't figure out where or how the `a` elements are being rendered? Would this change even be possible?

I'm fixing up a few other things which I'll send along in a PR, but this one's got me in a loop. Any insight to where this change would be made would be appreciated!

Ethan-Arrowood:
HI @svinkle could you inspect element on the exact elements you are referencing and post a screenshot here please? This will help me as I traverse the React structure.

svinkle:
Ah yes, that would have beeb helpful.

Here you go, @Ethan-Arrowood:

![screen shot 2017-10-15 at 11 06 49 am](https://user-images.githubusercontent.com/1392632/31586001-0c046d80-b199-11e7-91aa-7344577f411f.png)

Ethan-Arrowood:
The NavPanel is made up of individual React-Bootstrap [Panel](https://react-bootstrap.github.io/components.html#panels) elements. The <a> element is created by this component. I'd recommend investigating the [Panel](https://github.com/react-bootstrap/react-bootstrap/blob/master/src/Panel.js) source code. Try to find the parameter the <Panel /> component passes to the <a> href property. Once you figure that out, update the code in NavPanel.js where we are rendering the <Panel /> element. 

Give it a try, and if you can't find it I can help you more.

svinkle:
Got it! Good pointers, @Ethan-Arrowood. Thanks!

Ethan-Arrowood:
You should keep this issue open, edit the code, and submit a PR and reference this issue. Once your PR is merged we can close the issue. When you do submit a PR let me know I'd be happy to review it for you 👍

svinkle:
Ah yes, good flow there. Will do.

[...]

PR #4175, opened 2017-10-24
Added skip link

svinkle:
## What does this do?
This PR adds a "skip to main content" link which is a common feature to skip past repetitive content such as the header and side navigation areas for keyboard users.

## Change log
- Adds additional CSS to present the skip link in an appropriate manner
- Adds a "skip link" above the main logo
- Adds `role="main"` for landmark navigation along with `tabindex="-1"` in order for this element to receive keyboard focus
- Adds `role="banner"` to add additional landmark navigation

## Screens
Skip link only appears on keyboard `:focus`. Everything should look and behave the exact same for sighted mouse users. 

skip-link

## Related issues

Related to #897.

### DIFF
--- a/src/layouts/Layout.jsx
+++ b/src/layouts/Layout.jsx
@@ -22,10 +22,20 @@ function Layout(props) {
 
   return (
     <div>
-      <Grid fluid={ true }>
+      <Grid
+        fluid={ true }
+        id='header'
+        role='banner'
+        >
         <Row>
           <Navbar className='navBar'>
             <Col md={ 3 } xs={ 12 }>
+              <a
+                className='skip-link sr-only sr-only-focusable'
+                href='#main'
+                >
+                Skip to main content
+              </a>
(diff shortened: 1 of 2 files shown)


src/layouts/Layout.jsx
```diff
@@ -52,7 +62,13 @@ function Layout(props) {
           <Col md={ 4 }>
             <SideNav />
           </Col>
-          <Col className='content' md={ 8 }>
+          <Col
+            className='content'
+            id='main'
```
Bouncey: I was thinking about this the other day and this looks like a good opportunity to implement it

Can you wrap `{ props.children() }` in a `<main>` tag please whilst you are here. You can then transfer themain id and role tags to it if required.

svinkle:
@Bouncey `main` added. 👍

Bouncey:
Thanks @svinkle 

PR #4184, opened 2017-10-24
Erase–remove idiom

Nik-Pavlov:
It is good to know the best way to remove elements that fulfill a certain criterion from a container. And in my experience it is very popular interview question.


Nik-Pavlov:
I've done some refactoring of the article. Would you be so kind @erictleung, to take a look?

PR #4191, opened 2017-10-24
Create accessibility-examples.md

kelsS:
Started a document providing real world accessibility application examples along with descriptions and links for further information.


domwakeling:
Hey @kelsS I'd love to be able to merge this PR, but think it would be better placed if it was moved up to the `accessibility` folder rather than sitting as a sub-folder under `accessibility/accessibility-basics`.

If you can change that I'll be able to merge 👍 

It would be great if you could also add to the stubs for Tables and Tags, but not neccessary to get an initial merge. ✍️ 

Thanks for your work on this article 😄

PR #4247, opened 2017-10-24
Adding Multi-dimensional Arrays in C

davidhayden:


imcodingideas:
@davidhayden thank you for sharing your knowledge with C with the rest of the community. I know @luandy64 asked to add some more links but that would be an opportunity for another pull request so I will merge this in.

PR #4254, opened 2017-10-24
Update index.md

ywiley94:
JavaScript numbers description. With information pulled from multiple sources and summarized into the simplest terms.


johnkennedy9147:
@ywiley94 closing PR as it doesnt really improve on the existing article and you have added stub text back in and also removed a good link

PR #4270, opened 2017-10-24
Updated index.md with check # of factors

CodeFay:
Added line 54 onward to confirm we have identified total number of positive factors based on prime factorization tree.


ryanwhocodes:
@CodeFay Great contribution!

PR #4390, opened 2017-10-24
Update index.md

rnnr100:
Added Vim logo and linked logo to Vim homepage.


domwakeling:
There are a few articles which have logos (not many) - but suggest it needs to be at the head of the article rather than at the end.

PR #4401, opened 2017-10-24
Added User Flow Article

rickspm:


erictleung:
Blatant copy-paste from https://www.optimizely.com/optimization-glossary/user-flow/ without attribution nor re-phrasing. The website doesn't indicate re-sharing or copying of their materials, so I'll close this PR.

PR #4462, opened 2017-10-24
Added stub for perceptron under neural-networks

ysharc:
Wanted to added more info on neural networks, like perceptron, multi-layer perceptron, CNNs, RNNs etc., Adding this stub to make sure that it is the right directory structure.


Asjas:

PR #4495, opened 2017-10-25
add content to the PHP intro article

n8finch:
Hey FCC!

I wanted to add a little statistic on usage and external resources people could use for getting started. I feel that, because PHP has been so widely used for so long, it's important for folks to know how prevalent it is, not only with legacy work, but also with new development on frameworks like Laravel and Symfony. 

I also thought it was important for folks to have a list of external resources they could go to to start learning PHP if they wanted to. 

Hope that helps!
Nate


domwakeling:
Hi @n8finch. Congrats on your first pull request (PR)! 🎉

PR #4547, opened 2017-10-25
Explained Freecodecamp Certificates

fassousa:


Bouncey:
@fassousa Sorry it took so long to get looking at it. As you can see we still have over 2800 PR's to get through. Thank you for your patience

PR #4553, opened 2017-10-25
Added stubs for gradient descent and optimization algorithms for gradient descent

ysharc:
I think the guides are missing the important stuff of neural networks. So, added the stubs for gradient descent and optimization algorithms for gradient descent. Not sure, whether each of the optimization algorithms like Adam, Adagrad, rmsprop etc., should have their own sub-page or they go into the stub I just created.

### DIFF
--- a/src/pages/machine-learning/deep-learning/gradient-descent/index.md
+++ b/src/pages/machine-learning/deep-learning/gradient-descent/index.md
@@ -1,0 +1,15 @@
+---
+title: Gradient Descent
+---
+## Gradient Descent
+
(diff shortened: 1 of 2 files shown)


luandy64:
Hey @ysharc 

Since the Guides should serve as a introduction to the topics it contains, I think it's safe to just have introductions to Gradient Descent and Optimizations for Gradient Descent.

Then within the broad/general article, there can be links to other resources that explain more about particular algorithms.

Just my opinion, hopefully other people can chime in here

luandy64:
Hey @ysharc 

Since the Guides should serve as a introduction to the topics it contains, I think it's safe to just have introductions to Gradient Descent and Optimizations for Gradient Descent.

Then within the broad/general article, there can be links to other resources that explain more about particular algorithms.

Just my opinion, hopefully other people can chime in here.

PR #4601, opened 2017-10-25
Added stub for R language

ysharc:


erictleung:
Looks like you've pulled a lot of other commits with this one. Plus, it looks like there's already a stub created. Happy coding!

PR #4602, opened 2017-10-25
Article content added

jeevan-jp:
I found [this](https://github.com/freeCodeCamp/guide-articles/blob/master/articles/CSS/css-flexbox-tips-and-tricks/index.md) article of @Bouncey good. It's better to keep things explained rather than adding references only. However, external links can be provided at the of article. 
(Not meant to criticize anyone)
Thank You.

### DIFF
--- a/src/pages/cplusplus/if-else-statement/index.md
+++ b/src/pages/cplusplus/if-else-statement/index.md
@@ -122,6 +122,35 @@ You entered a negative number : -200
 This step is always printed
 ```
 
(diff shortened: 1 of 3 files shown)


raisedadead:
Hi @jeevan-jp 

Thank you for taking the time to create this Pull Request.

It seems that an edit for this article has already been accepted. If you feel you could add to the current article, please open a new Pull request

Happy Contributing!

PR #4624, opened 2017-10-25
added my changes

prabusubra:
updated about nested ternary operator


prabusubra:
updated about nested ternary operator

PR #4655, opened 2017-10-25
Opening a file in a specific line number

wbox:


ryanwhocodes:
@wbox Thank you for your patience and contribution to the freeCodeCamp guide!

PR #4671, opened 2017-10-25
improve the look of search results

mrcrmn:
It gives the resultIcon a fixed width which aligns the following text.
It also adds a little margin to each result element.

It looks nicer in general

### DIFF
--- a/src/css/main.scss
+++ b/src/css/main.scss
@@ -197,6 +197,10 @@
   width: 100%;
 }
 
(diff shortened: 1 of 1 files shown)


QuincyLarson:
@mrcrmn Thanks for your patience. Our search results code has shifted quite a bit since you opened this, so I'm closing this. We'd welcome future contributions, though, and will review them much faster.

PR #4696, opened 2017-10-25
corrections and further explanation

toby2all:
Added more explanation to further clarify the concept of CSS classes and cleaned up the example code as well.


QuincyLarson:
@toby2all Thanks!

PR #4710, opened 2017-10-25
Reformat C++ Input and Output code samples

anthonylewis:
Apply consistent spacing to code samples. Clean up prose a bit. Add a (hopefully) helpful tip about visualizing the flow of data when using stream operators.


anthonylewis:
Hey @luandy64 this should be good to go now.

PR #4712, opened 2017-10-25
Removing instructions and references to dead link

jcoulter:
I searched for a copy of the animated gif but wasn't able to find it. Therefore, I deleted the reference and the dead link.


QuincyLarson:
@jcoulter Thanks for your extreme patience. This got buried during Hacktoberfest last year. Your edits look great, though we've already gone through and updated the copy since then.

PR #4715, opened 2017-10-25
Django-Overview

YatinGupta777:
Gives a brief point of why django is used as a web framework


src/pages/python/web-frameworks-and-what-they-do-for-you/Django-Overview
```diff
@@ -0,0 +1,21 @@
+title:Django
```
bla-ke: Change your first line:
```
---
title: Django
---
```

PR #4717, opened 2017-10-25
Added an image of Bootstrap carousel

mrMarce:
Added a graphic for visual representation to aid in learner understanding.


systimotic:
Hi @mrMarce! Can you add the example in code instead of an image? That way, it'll not only look the way it should, but also work the way it should.

Happy coding!

PR #4720, opened 2017-10-25
For-Loop

YatinGupta777:
Added the basics of using a for-loop in c++


src/pages/cplusplus/For-Loop
```diff
@@ -0,0 +1,39 @@
+title:for-loop
```
bla-ke: Change this first line to:
```
---
title: For Loop
---
```

PR #4728, opened 2017-10-25
Added an example image of a Bootstrap Dropdown

mrMarce:


systimotic:
Hi @mrMarce! Can you add the example in code instead of an image? That way, it'll not only look the way it should, but also work the way it should.

Happy coding!

PR #4733, opened 2017-10-25
Added image example of Bootstrap Buttons

mrMarce:


systimotic:
Thanks for contributing @mrMarce! As we already have the actual buttons on the page, I think adding an image does not add value.

Happy coding!

PR #4738, opened 2017-10-25
please check out added short definiton

randomguy420:
a short version of lambda function


dhcodes:
Thanks for the contribution @randomguy420

PR #4793, opened 2017-10-25
Add text"hardly useful for the masses"

adidtu27:
To show a negative aspect of irresponsible code development.


IsaacAbrahamson:
@Ethan-Arrowood, I marked this as invalid, I would like a second review.

PR #1614, opened 2017-10-19
grammar correction - "you" to "you're"

ajwallacemusic:

PR #4838, opened 2017-10-25
Fix grammar

nwarwick:


nwarwick:
@Ethan-Arrowood Is there a resource specifying the minimum amount of changes for a commit to be merged? 
#1614 This change would be invalid as well.

PR #4839, opened 2017-10-25
Added how to use virtual environments with Python

oribaldi:
Includes:
- Short explanation of what are virtual environments and why are they useful.
- Introduction to virtualenv and Conda.


ryanwhocodes:
@oribaldi Thank your for your patience! This is a solid article!

PR #4932, opened 2017-10-25
Normalise Articles 25Oct17

Bouncey:


src/pages/data-science-tools/jupyter-notebook/index.md
```diff
@@ -10,4 +10,6 @@ This is a stub. <a href='https://github.com/freecodecamp/guides/tree/master/src/
 <!-- The article goes here, in GitHub-flavored Markdown. Feel free to add YouTube videos, images, and CodePen/JSBin embeds  -->
 
 #### More Information:
-[Jupyter/IPython Notebook Quick Start Guide](http://jupyter-notebook-beginner-guide.readthedocs.io/en/latest/what_is_jupyter.html)
+<!-- Please add any articles you think might be helpful to read before writing the article -->
```
HKuz: Hmmm, looks like there was only a link added to this stub, that the Normalise script removes. Should we add it back, or will this keep happening until the stub language is replaced with an article?

src/pages/algorithms/avl-trees/index.md
```diff
@@ -1,7 +1,7 @@
 ---
-title: AVL Trees
+title: Avl Trees
```
HKuz: This changes `AVL` to `Avl` - do we need to update a proper nouns list or somehow mark that the all-caps version is the right way to display this?

src/pages/csharp/xaml/index.md
```diff
@@ -1,6 +1,15 @@
 ---
-title: XAML
+title: Xaml
 ---
```
HKuz: Same comment as the `AVL` title re capitalization 😄

Bouncey:
@HKuz all three review note are linked to the same issue I had with this. 

Normalise looks for `'This is a stub...` in the file via a RegEx. If found, the entire file is re-stubbed. This behaviour was intended, allowing us to update all stubs if the template stub changed for any reason. 

The articles on AVL and XAML, and five others, still had stubs in the file and were re-written. I managed to catch these as the diff was quite obvious. 

Thank you for taking the time to QA this as thoroughly as you have, I'll make the changes shortly.

HKuz:
# LGTM!! 🤙 ☎️ 🎵 

Thanks @Bouncey - that's good to know about the stub being re-written so we don't just merge in added links without some form of an article to go with it 👍  I've been meaning to dig into the scripts more to better understand the site, maybe a weekend activity while reviewing a couple hundred PRs or so... 😆

PR #4937, opened 2017-10-26
Update javascript slideshow tutorial with content

andres-uris:
This is a first simple go to the tutorial with just one functionality that is switching image in a time interval.


Asjas:

PR #4939, opened 2017-10-26
Added more information on scopes

nfayazov:
I think the existing text could be more concise, but if not changed, I think this commit has information that is crucial for a full understanding of JavaScript scope.


Asjas:

PR #4992, opened 2017-10-26
wrote an article on absolute functions

Ray6464:


Ray6464:
Thanks mate, looking forward to contributing more!

PR #5004, opened 2017-10-26
add latest version

c8001221:


systimotic:
Hi @c8001221!

Thank you for contributing. You have created four PR's within half an hour. I think the content is very good. There's a problem, though: they all edit the same file. When we merge your PR's, git will not know how to properly to put them in the right order. I'm closing these four PR's. Can you please take the great content you wrote for them and create a single PR that does it all at once? Thanks!

Happy coding!

PR #5047, opened 2017-10-26
Updated semi-colons for javasctipt

nrum:


tchaffee:
@nrum The code in this PR is not formatted correctly. Could you fix that so we can review and possibly merge?

PR #5108, opened 2017-10-26
added information regarding ASP.NET

khambley:
Added information regarding the .NET framework and how that works with C#. I want to add a topic called 'ASPNET' as a folder and write about the framework in more detail, then add a link back here to the ASPNET page. With the open sourcing of .NET code in 2008 and .NET applications becoming machine independent, it has become a relevant platform, especially in business use cases, so it should at least get some airtime here.


scissorsneedfoodtoo:
@khambley, everything LGTM! Thank you for helping us improve this section of the guide. Looking forward to your next contribution! :+1: :+1:

PR #5118, opened 2017-10-26
Update index.md

Pranav-Sharma-23:
I initialised a variable with a different value.


tchaffee:
@Pranav-Sharma-23 It's not clear why you used a different value? But if you want to use a different value, you also need to update the code explanation, which still uses the old value.

PR #5137, opened 2017-10-26
Search results announcements

svinkle:
## What does this do?
This PR adds `aria-live` regions to the search templates in order to announce the result count for screen reader users.

## Change log
- Adds new, visually hidden `div` with `role="status"` and `aria-live="polite"` to `NoResults.jsx` in order to convey "no results" search status.
- Adds new, visually hidden `div` with `role="status"`, `aria-live="polite"`, and `aria-atomic="true"` to `Results.jsx` in order to convey current number of search results.
- Adjusts logic to ensure live region is always available.

## Screens
Everything should look and behave the exact same for sighted mouse users.

screen shot 2017-10-26 at 12 51 52 pm

## Related issues

Related to #897.

### DIFF
--- a/src/pageComponents/search/NoResults.jsx
+++ b/src/pageComponents/search/NoResults.jsx
@@ -24,6 +24,13 @@ function NoResults(props) {
           </Button>
         </Link>
       </div>
+      <div
+        aria-live='polite'
+        className='sr-only'
+        role='status'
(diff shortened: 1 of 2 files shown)


Bouncey:
## LGTM :+1: 

Thanks @svinkle 

PR #5160, opened 2017-10-26
Some information about JavaScript.

avijit1998:


scissorsneedfoodtoo:
@avijit1998, thank you for taking the time to improve this section of the guide. Unfortunately, this stub has been updated by another PR that was accepted earlier, so I'll have to close this. Also, I believe this was taken directly from the MDN site without modification. Please feel free to help us improve other sections of the guide, but remember that the content submitted has to be original / modified and not taken directly from another source. Happy coding! :+1:

Issue #1297, opened 2017-10-19
Dynamic meta data for Twitter and Facebook cards


QuincyLarson:
Currently Guide articles don't get their own cards on Twitter or Facebook. 

We want to turn raw URLs like this:

![quincy_larson_on_twitter___test_article_https___t_co_acac3vne2i_](https://user-images.githubusercontent.com/985197/31762147-ed0b530a-b47f-11e7-9671-161b3ec3ca17.png)

Into nice looking cards like this:

![quincy_larson_on_twitter___test_https___t_co_a4uwqx6gfz_](https://user-images.githubusercontent.com/985197/31762201-1650f954-b480-11e7-8c2e-23cbf698a0d5.png)

This involves creating meta tags, like the ones we use here: https://github.com/freeCodeCamp/donate/blob/master/index.html

We would want to dynamically update the article title and other meta information based on the individual article.

@systimotic @Bouncey Do you have any thoughts on the best way to do this?

LulaV14:
Hi @QuincyLarson,  may I help with this?

soueuls:
I am interested to work on this with @LulaV14 if she is in!

I made some research and twitter partially support the OpenGraph protocol as a fallback (we should favor it over the native twitter protocol this way we get twitter/facebook without too much code duplication.

```
<meta name="twitter:card" content="summary" />
<meta name="twitter:site" content="@freeCodeCamp" />
<meta name="twitter:creator" content="X" />
<meta property="og:url" content="X" />
<meta property="og:title" content="X" />
<meta property="og:description" content="X" />
<meta property="og:image" content="X" />
```
As you can see the two first meta tags can be set statically.

- twitter:creator we could allow the writer to add an author tag in the article front matter (and have FreeCodeCamp as a fallback) or simply recognize the fully open source nature of these guides and statically set the author to be FreeCodeCamp.
- og:url is trivial to set dynamically.
- og:title we could use the title set in the article front matter (but then we lose some context), the title of javascript's break statement is simply called **break statement*. Thus, I propose to use the breadcrumb smartly concatenating the first (giving the general context) with the last (giving the specifics)
- og:description could be the first paragraph?
- og:image we could build a script fetching the first image of the article matching a decent twitter/facebook ration.

I am curious to get your feedbacks, let's discuss and tackle this 🤓

QuincyLarson:
[...]

PR #5225, opened 2017-10-26
Feat: Dynamic meta data for Twitter and Facebook cards

LulaV14:
**Description**
I'm creating this PR to fix issue #1297 

Although is not completely done, I'll be working with @soueuls to finish this fix.

### DIFF
--- a/src/templates/Article.jsx
+++ b/src/templates/Article.jsx
@@ -5,6 +5,8 @@ import Helmet from 'react-helmet';
 
 import Breadcrumbs from '../templateComponents/Breadcrumbs.jsx';
 
+import titleify from '../../utils/titleify';
+
 const propTypes = {
   data: PropTypes.object,
   location: PropTypes.object
@@ -22,6 +24,21 @@ function Article(props) {
       title
     }
   } = article;
+
+  const pathMap = pathname
+    // remove leading and trailing slash
+    .replace(/^\/([a-z0-9/-]+[^/])\/?$/i, '$1')
+    .split('/');
+
+  var metaTitle = '';
+  pathMap.forEach((title, i) => {
+    if (i === pathMap.length - 1) {
+      metaTitle += titleify(title);
+    } else {
+      metaTitle += titleify(title + ' | ');
+    }
+  });
+
   return (
     <div>
       <Helmet>
@@ -29,6 +46,14 @@ function Article(props) {
         <link
           href={ `https://guide.freecodecamp.org${slug}` }
           rel='canonical'
+        />
+        <meta
+          content={ `https://guide.freecodecamp.org${slug}` }
+          property='og:url'
+        />
+        <meta
+          content={ `${metaTitle}` }
+          property='og:title'
         />
       </Helmet>
       <Breadcrumbs path={ pathname } />
--- a/src/templates/Article.jsx
+++ b/src/templates/Article.jsx
@@ -12,6 +12,47 @@ const propTypes = {
   location: PropTypes.object
 };
 
+function getOgTitle(pathname) {
+  const pathMap = pathname
+    // remove leading and trailing slash
+    .replace(/^\/([a-z0-9/-]+[^/])\/?$/i, '$1')
+    .split('/');
+
+  let metaTitle = '';
+  pathMap.forEach((title, i) => {
+    if (i === pathMap.length - 1) {
+      metaTitle += titleify(title);
+    } else {
+      metaTitle += titleify(title + ' | ');
+    }
+  });
+
+  return metaTitle;
+}
+
+function getOgDescription(html) {
+  const el = document.createElement('html');
+  el.innerHTML = html;
+  let description = '';
+  const paragraph = el.getElementsByTagName('p')[0];
+  if (paragraph) {
+    description += paragraph.innerText;
+  }
+  return description;
+}
+
+function getOgImage(html) {
+  const el = document.createElement('html');
+  el.innerHTML = html;
+  const image = el.getElementsByTagName('img')[0];
+  if (image) {
+    return image.src;
+  } else {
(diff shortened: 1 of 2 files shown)


src/templates/Article.jsx
```diff
@@ -22,6 +24,21 @@ function Article(props) {
       title
     }
   } = article;
+
+  const pathMap = pathname
+    // remove leading and trailing slash
+    .replace(/^\/([a-z0-9/-]+[^/])\/?$/i, '$1')
+    .split('/');
+
+  var metaTitle = '';
```
Bouncey: Why can't we use `title` from the `article`? On line 24

Bouncey: Please declare with either `const` or `let`

Bouncey:
Changes to `meta.js` require a stop/start of gatsby as it effects the html template. This is maybe why you didn't see the changes immediately

src/templates/Article.jsx
soueuls: I can't check the value of `title` right now (but if it's based on the page title or the front-matter specified one) we are afraid that it's going to be contextless is some cases.

Here is an [example](https://guide.freecodecamp.org/cplusplus/input_and_output) where the title is `Input and Output` without any mention of the context being C++.

We are going to work on a strategy to build it up from the url path, I think @LulaV14 is close to a solution.

soueuls:
@LulaV14 I made some research and I either need to make some pull request on your fork or you can add me as a collaborator with write access to it

LulaV14:
@soueuls I've added you as a collaborator 😊 

Were you able to run the project so you can check my changes?
For example with this [url](https://guide.freecodecamp.org/java/loops/for-loop) `title` is "For Loop" and  `metaTitle` is "Java | Loops | For Loop" 

The problem is that even though by inspecting the elements you can see the tag 
`<meta content="Java | Loops | For Loop" property="og:title" data-react-helmet="true">`
using the [twitter card validator](https://cards-dev.twitter.com/validator) with ngrok I get an error.
This is the log I get:
```
INFO:  Page fetched successfully
INFO:  8 metatags were found
INFO:  twitter:card = summary tag found
ERROR: Possibly invalid value (if specified) (og:title)
ERROR: Possibly invalid value (if specified) (twitter:text:title)
ERROR: Possibly invalid value (if specified) (twitter:title)
```

soueuls:
@LulaV14 That's also the first time I am contributing with someone else on a repository I have no write access to, so I added your repository as upstream to work on top of your initial pull request. Made some changes and then made a pull request to freeCodeCamp. (We would need help to know if that's indeed the correct workflow)

My changes implement all the missing OG properties, fix the let/const.
I think what we are doing is wrong, we inject the title, description, and image dynamically but twitter reads the file content statically, therefore, the tags are not here yet.

Bouncey:
@soueuls that sounds about right, in development, everything about the article is dynamic. You would have to build and host the site to check that twitter is seeing everything properly.

QuincyLarson:
@LulaV14 This looks great. Were you able to address the changes @Bouncey requested? We're eager to get this live so these articles will look better when shared :)

LulaV14:
Hi @QuincyLarson! Thanks, I've added the changes from @soueuls and deleted the cheerio implementation as requested. But I'm having trouble with the deployment to be able to test that everything is displaying properly.
What I'm doing is run `yarn run build` and it builds correctly but when I try to deploy to heroku it doesn't work using the node buildpack nor the react buildpack. Can you please help me with that?

Bouncey:
@LulaV14 we deploy with netlify, you should be able to host this branch there and test

src/templates/Article.jsx
```diff
@@ -5,11 +5,54 @@ import Helmet from 'react-helmet';
...
+    } else {
+      metaTitle += titleify(title + ' | ');
+    }
+  });
+
+  return metaTitle;
+}
+
+function getOgDescription(html) {
+  const el = document.createElement('html');
```
Bouncey: We should be able to get an extract from graphql. Copy the page query at the end of the file and have a play around with the graphql server is which is available in development.

src/templates/Article.jsx
```diff
@@ -5,11 +5,54 @@ import Helmet from 'react-helmet';
...
+  let description = '';
+  const paragraph = el.getElementsByTagName('p')[0];
+  if (paragraph) {
+    description += paragraph.innerText;
+  }
+  return description;
+}
+
+function getOgImage(html) {
+  const el = document.createElement('html');
```
Bouncey: I would query `markdownRemark` for this in graphql. `remark` converts the markdown file in to an AST. This might still be hanging around

LulaV14:
@Bouncey I did try to use GraphQL as you requested but wasn't able to get anything from the html there. It may be done if we put fields for description and image just like it's done for the title in each post 
```
---
title: Tables
description: Tables description
image: https://somewebsite.com/images/tables.jpg
---
```

But since I wasn't successful with GraphQL what I did was to extract from the html variable the first paragraph and the first image using regex. I tested in local and also deployed to Netlify to be able to test it with the [twitter card validator](https://cards-dev.twitter.com/validator) and it works 😊  

Here is how it looks on Facebook:
screen shot 2017-11-13 at 6 21 28 pm

QuincyLarson:
@LulaV14 Thanks for your extreme patience. This looks great! This is a huge improvement over what we have now. I'm going to wait for @Bouncey to take a quick look at this and verify this is ready to merge.

Bouncey:
# LGTM

Thanks @LulaV14 and @soueuls :+1:

QuincyLarson:
@LulaV14 Thanks again for fixing this!

LulaV14:
@QuincyLarson @Bouncey Thanks for letting me contribute to this awesome repo 😄 !! 
'Will be glad to be of any help again 😊

PR #5228, opened 2017-10-26
Updated HelloWorld.java

Tushcr7:
**Description**


jamesjpk123:
Seems like a trivial addition. I don't think adding exclamations to "Hello World" can increase the learning of java in readers.

PR #5243, opened 2017-10-26
changed "etc" to "etc."

deguzmankevin:
**Description**


jamesjpk123:
Looks like a trivial edit, only adding a period. Can you do more to improve the article @deguzmankevin?

PR #5245, opened 2017-10-26
Added my font property changes

bflat96:
**Description**

### DIFF
--- a/src/pages/css/properties/font-property/index.md
+++ b/src/pages/css/properties/font-property/index.md
@@ -3,13 +3,8 @@ title: Font Property
 ---
 ## Font Property
 
-This is a stub. <a href='https://github.com/freecodecamp/guides/tree/master/src/pages/css/properties/font-property/index.md' target='_blank' rel='nofollow'>Help our community expand it</a>.
+My Font Property Changes
 
-<a href='https://github.com/freecodecamp/guides/blob/master/README.md' target='_blank' rel='nofollow'>This quick style guide will help ensure your pull request gets accepted</a>.
+The font property define the font family, boldness, size, and the style of a text. It has two types of font family names:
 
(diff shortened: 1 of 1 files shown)


src/pages/css/properties/font-property/index.md
```diff
@@ -3,13 +3,8 @@ title: Font Property
...
 
-This is a stub. <a href='https://github.com/freecodecamp/guides/tree/master/src/pages/css/properties/font-property/index.md' target='_blank' rel='nofollow'>Help our community expand it</a>.
-
-<a href='https://github.com/freecodecamp/guides/blob/master/README.md' target='_blank' rel='nofollow'>This quick style guide will help ensure your pull request gets accepted</a>.
-
-<!-- The article goes here, in GitHub-flavored Markdown. Feel free to add YouTube videos, images, and CodePen/JSBin embeds  -->
-
-#### More Information:
-<!-- Please add any articles you think might be helpful to read before writing the article -->
+My Font Property Changes
```
jamesjpk123: This line is not needed.

src/pages/css/properties/font-property/index.md
```diff
@@ -3,13 +3,8 @@ title: Font Property
...
 
+The font property define the font family, boldness, size, and the style of a text. It has two types of font family names:
```
jamesjpk123: Can you elaborate more on how the property "has two types of font family names"? Can you add in more information about how it can be used? Try looking at https://www.w3schools.com/cssref/pr_font_font.asp for help.

bflat96: The font family has two types to distinguish between what other fonts are available to the browser. The font you want to choose would be the first font listed in the font-family and end with the generic family as the fallback.

PR #5263, opened 2017-10-27
Added the paragraph.

arkaseth:
**Description**


QuincyLarson:
It looks like some of this was copied from a wikipedia article. Always cite your sources.

PR #5284, opened 2017-10-27
Added comment

FieryNyan:
####Nicely written! Was wondering if we can create a sub-list inside a sub-set? Like how it's possible to create outlines in google docs?

**Description**


ryanwhocodes:
@FieryNyan Thank you for your comment. If you'd like to make a change in the freeCodeCamp guide, please submit a PR with your addition. Alternatively, you could discuss it in the freeCodeCamp forum at https://www.freecodecamp.org/forum/

PR #5296, opened 2017-10-27
Create Array Destructuring Guide

neel2911:
**Description**


scissorsneedfoodtoo:
@neel2911, thank you for taking the time to improve this section of the guide. Unfortunately, too much of this seems to come directly from the MDN article, often word-for-word. Plagiarism isn't allowed in the guide, so would you be able to rename the sections and reword a lot of the sentences? Some examples are directly from the MDN article, too, so those will also need to be reworked.

After that, I'd be happy to take another look at your work before we merge your PR. Thanks!

PR #5299, opened 2017-10-27
Added Conditional Statements

ShadowDLL:
Conditional Statements ( IF / ELSEIF / ELSE )

**Description**


scissorsneedfoodtoo:
@ShadowDLL, thank you for taking the time to improve this section of the guide. Unfortunately, a lot of this seems to be taken from the w3schools article. Plagiarism isn't allowed in the guide, so you'll have to rework a lot of these sentences and examples. I'd be happy to take another look at your PR after you do that!

PR #5335, opened 2017-10-27
Made some changes.

rbrishabh:
Fixed typos.

**Description**


domwakeling:
Hey @rbrishabh . I think this reads better with “Finally” (ie “this is the final option that we will discuss”) rather than “Eventually”.

However I’d like to thank you for your PR and look forward to your contributions in the future

PR #5371, opened 2017-10-27
Added new article for php cookies

Kamalhb:
**Description**


scissorsneedfoodtoo:
@Kamalhb, thank you for taking the time to improve this section of the guide. Unfortunately, this stub has been updated by another PR that was accepted earlier, so I'll have to close this. Also, a lot of this seems to be taken directly from https://www.tutorialspoint.com/php/php_cookies.htm. Plagiarism is not allowed in the guide, so please keep that in mind in the future.

PR #5407, opened 2017-10-27
Added Sources heading and two citation/sources

SiriusScaper:
I noticed that some of this article (beginning paragraph, code, and second section) is pretty directly taken from two sources that I had previously read. According to the style guide, I believe they need to be cited properly.

### DIFF
--- a/src/pages/javascript/promises/index.md
+++ b/src/pages/javascript/promises/index.md
@@ -2,7 +2,7 @@
 title: Promises
 ---
 ## Promises
-JavaScript is single threaded, meaning that two bits of script cannot run at the same time; they have to run one after another. A Promise is an object that represents the eventual completion (or failure) of an asynchronous operation, and its resulting value.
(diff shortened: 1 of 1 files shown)


scissorsneedfoodtoo:
@SiriusScaper, thank you for your contribution. Great catch! Definitely not good that some plagiarized material got through. Unfortunately, there seems to be a conflicting file here. Also, I'm thinking that it's probably better if we can reword a lot of these sentences, rather than add sources. Since many of them directly from the sources you listed, they would have to be direct quotes, quote marks and all. The whole article would just be a mess punctuation wise.

Would you be interested in helping us with the rewrites?

SiriusScaper:
I could give it a go. A year later... :smile:

QuincyLarson:
Thank you for your patience.

We have archived this Guide repository and moved the guide into our main https://github.com/freecodecamp/freecodecamp repostory.

Since this Guide repository is now archived, I'm closing this pull request.

I encourage you to take a look at the repository and consider contributing to it this Hacktoberfest.

We now have a much larger team QA'ing these pull requests and we should be able to review your pull request much faster this time. Here's an article I wrote with details for how to do this: https://www.freecodecamp.org/n/FDoftlSup

Thank you again for contributing to open source.

PR #5420, opened 2017-10-27
Update index.md

andrea1015:
**Description**


scissorsneedfoodtoo:
@andrea1015, thank you for taking the time to improve this section of the guide. However, a lot of this seems to be taken directly from https://en.wikipedia.org/wiki/Multiplication. Plagiarism is not allowed in the guide, so you'll have to rewrite this. Please let us know when you're finished and someone will be happy to take another look at your PR.

PR #5464, opened 2017-10-27
Fix code block to display correctly in VIM Macros Page

aribeiro:
In the code block bellow "To start a Macro, in Normal Mode, press:" the commnad shoud be `q<letter>`, but in the guide page, it renders as an html tag `<letter><letter/>` and it is not visible for the visitors, that only see `q`.
So I add an identifier on the block, and change the description, in the hope it displays correct in the guide.

**Description**


domwakeling:
Hey @aribeiro Thank you for your contribution to the page! 👍
We're happy to accept these changes, and look forward to future contributions. 🎉

PR #5469, opened 2017-10-27
info about text formatting html updated

Kyanij:
**Description**


scissorsneedfoodtoo:
@Kyanij, thank you for taking the time to improve this section of the guide. Unfortunately, this stub has been updated by another PR that was accepted earlier, so I'll have to close this. Also, plagiarism is not allowed in the guide, so please keep that in mind for the future.

PR #5485, opened 2017-10-27
weird if statement

ishantgarg40:
**Description**


Bouncey:
@ishantgarg40 There is no context to your changes, I don't know what it adds to the article. Why have you placed it just after the comparison operators?

PR #5518, opened 2017-10-27
JavaScript: String.prototype.endsWith

ZaninDavide:
Wrote the article about String.prototype.endsWith


ryanwhocodes:
@ZaninDavide Thank you for your patience and for another useful contribution to the freeCodeCamp guide!

PR #5543, opened 2017-10-27
Updated "Why do you need a framework?"

Bakley:
**Description**


scissorsneedfoodtoo:
@Bakley, thank you for taking the time to improve this section of the guide. However, a lot of this seems to be taken directly from http://www.softkul.com/2018/03/facts-you-must-know-about-django-python.html. Plagiarism is not allowed in the guide, so you'll have to rewrite this. Please let us know when you're finished and someone will be happy to take another look at your PR.

PR #5569, opened 2017-10-27
Add a Why you should use section

armandfardeau:
It seems to me good to explain to the beginner the interest of something so that he understands the ins and outs of it.

**Description**


ryanwhocodes:
@armandfardeau Thank you for your patience! This is a useful contribution. Congratulations on your first PR merged into the freeCodeCamp guide!

PR #5658, opened 2017-10-27
Write article about hexadecimal numbers,

64Mega:
**Description**
Wrote a comprehensive article on hexadecimal numbers and their uses in JavaScript and CSS.


64Mega:
@tchaffee I made that PR nearly a year ago, so my memory of my original intent with the article is a little rusty. Resolved the conflict as best I could.

PR #5901, opened 2017-10-28
Added article on Ddos

sammielas:
**Description**


1andee:
Hey @sammielas ⚠️

You plagiarized parts of this article, which goes against our [Academic Honesty policy](https://www.freecodecamp.org/academic-honesty). If you think that all the content can be safely taken from the following sources, we still need to cite them properly:

- https://www.cybrary.it/0p3n/dos-and-ddos/

Please read our [Contributing guidelines](https://github.com/freeCodeCamp/guides/blob/master/CONTRIBUTING.md) to see what we look for in a Pull Request.

PR #6025, opened 2017-10-28
Feat/feature twitter meta data

soueuls:
Collaboration with #5225

### DIFF
--- a/src/templates/Article.jsx
+++ b/src/templates/Article.jsx
@@ -5,6 +5,8 @@ import Helmet from 'react-helmet';
 
 import Breadcrumbs from '../templateComponents/Breadcrumbs.jsx';
 
+import titleify from '../../utils/titleify';
+
 const propTypes = {
   data: PropTypes.object,
   location: PropTypes.object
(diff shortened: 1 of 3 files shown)


src/templates/Article.jsx
```diff
@@ -2,14 +2,50 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import Helmet from 'react-helmet';
+import cheerio from 'cheerio';
```
Bouncey: It would be crazy to ship a 400k lib to each client just to populate meta tags.

This could be done during the build with a plugin, like we have with the forum emoji [here](https://github.com/freeCodeCamp/guides/tree/master/plugins)

soueuls: I have not worked with Gatsby in the past but I assumed `npm run build` would just generate all the html files needed to build the final website.

Therefore, the final html files would not have any reference to cheerio, right?

If I am wrong and it's not the case we can think of a way to effectively traverse the html content of articles.

Bouncey: Only just seeing this sorry. In answer to your question, yes gatsby generates static HTML files for each page, but at the client, it is just a react app.

Bouncey:
Is this or still required?

PR #6060, opened 2017-10-28
Add article for Math Ceil

vitsenklavitsen:
**Description**
Create Math Ceil article from scratch.


domwakeling:
Thanks again both, more contributions very welcome! 👍

PR #6090, opened 2017-10-28
Add article for Math Floor

vitsenklavitsen:
Add article for Math Floor from scratch

**Description**
Create article for Math Floor from scratch


vitsenklavitsen:
@luandy64 You're right. I changed it and I hope that it is a bit clearer now. Thanks for reviewing!

domwakeling:
Thanks both for your contributions to the page! 👍
We're happy to accept these changes, and look forward to future contributions. 🎉

PR #6159, opened 2017-10-29
Fixes rails new commands

magikid:
The `rails new app_name` commands in this file weren't rendering correctly.  I suspect that they were being filtered out as HTML since the were surrounded by greater-than and less-than symbols.

This updates the `rails new` commands to surround the application name with square brackets instead of the greater-than and less-than symbols.

**Description**


domwakeling:
Hey @magikid

It seems that similar changes have already been accepted earlier for this article you're editing, sorry about that. 😓

If you feel you have more to add, please feel free to open up a new PR.

Thanks again! 😊

---

> If you have any questions, feel free to reach out by commenting below. 💬

PR #6239, opened 2017-10-29
Added alignment options

BreakfastWithMe:
Inputted alignment codes examples and how we can use it on other attributes like images and not only for paragraphs.

**Description**


beaucarnes:
@BreakfastWithMe Thanks for your contribution. This has already been updated so I'll close this PR.

PR #6240, opened 2017-10-29
Update index.md

chandar18:
In case of OR if both operands are true then first value is returned and if both operands are false then second value is returned

**Description**


beaucarnes:
@chandar18 Looks good! Thanks for this PR.

PR #6292, opened 2017-10-29
Added basic description for accessible SVGs

StepanNaryshkov:
**Description**


StepanNaryshkov:
The charset attribute specifies the type of character-set encoding that this page uses. You should always define the character set.
Because rather than letting the browser auto-detect the charset from the code, we can supply it directly within the code. So occasions where browser making incorrect charset detections can be avoided.
However, this is particularly important if your page needs to display nonstandard characters.

`<meta charset="utf-8" />`  and `<meta http-equiv="Content-Type" content="text/html; charset=utf-8" />`
**In HTML5, they are equivalent**. Use the shorter one, it is easier to remember and type. [Browser support](https://code.google.com/archive/p/doctype-mirror/wikis/MetaCharsetAttribute.wiki) is fine since it was designed for backwards compatibility.
[more details](https://www.w3.org/International/questions/qa-html-encoding-declarations)

PR #6312, opened 2017-10-29
Developer tools: add Let's Encrypt stub

CanarySec:
**Description**
Add stub for Certbot article


beaucarnes:
@CanarySec Thanks!

PR #6317, opened 2017-10-29
Developer tools: add Certbot stub

CanarySec:
**Description**
add Certbot stub


beaucarnes:
@CanarySec Thanks. Good addition.

PR #6318, opened 2017-10-29
Security: add VPN stub

CanarySec:
**Description**
add VPN stub


beaucarnes:
@CanarySec Thanks. Good addition.

PR #6346, opened 2017-10-29
Java: Add details about Methods

Manish-Giri:
**Description**

Current guide on Methods doesn't talk about the various parts of a method body. Added a list breakdown that describes the different parts.


beaucarnes:
Looks good! Thanks for your contribution to the guide.

PR #6408, opened 2017-10-29
Remote vs onsite: multiple teams

geraintnyc:
Added description of multiple (independent) teams.


beaucarnes:
@geraintnyc Thanks!

PR #6446, opened 2017-10-29
Changed "OnPres" to "OnPress" in Button component

vtisnado:
The example won't run if you keep "OnPres" instead of "OnPress" and could confuse new ones.


domwakeling:
Hey @vtisnado

PR #6475, opened 2017-10-29
Update index.md

alenros:
elaborated on the use of else-if statments with an example

**Description**


beaucarnes:
@alenros Thanks!

PR #6479, opened 2017-10-29
Remove a trailing ")" on the last link

tschernobyl:
**Description**


erictleung:
@beaucarnes this is an easy fix. The Wikipedia has already been fixed, but there is still a trailing parens at the end.

All that needs to be done is resolve the conflict with adding in the correct Wikipedia link found in the current file.

Current file: https://github.com/freeCodeCamp/guide/blob/master/src/pages/security/backdoors/index.md

beaucarnes:
@erictleung Thanks for the info. I made the fix.
@tschernobyl Rest of PR looks good. Thanks for your contribution.

PR #6480, opened 2017-10-29
Added a short article about css floats.

focuse:
**Description**


raisedadead:
Hi @focuse 

Thank you for taking the time to create this Pull Request.

It seems that an edit for this article has already been accepted. If you feel you could add to the current article, please open a new Pull request

Happy Contributing!

PR #6485, opened 2017-10-29
Comparing null and undefined

TawkirA:
Comparing null and undefined to show how they behave differently.

**Description**


beaucarnes:
@TawkirA Looks good! Thanks for your contribution to the guide.

PR #6537, opened 2017-10-29
Improved the Vector Article C++

Thapz123:
I believed that looping over vectors is a pretty interesting tidbit that was necessary.

**Description**


beaucarnes:
Looks good! Thanks for your contribution to the guide.

PR #6562, opened 2017-10-29
Added description of storyboard

heatherdidit:
Defined what storyboards are in UX, how they work and added an example

**Description**


QuincyLarson:
@heatherdidit Thanks for your patience. Since the time that you created this PR, it looks like someone has gone through and created a comprehensive article on this.

PR #6563, opened 2017-10-29
Add CCS3 Media Query Information

carpenterca:
**Description**
I added basic instructions on how to use media queries as well as their uses with responsive web design concepts.


domwakeling:
Marked as stale and closed due to inactivity. :disappointed:

Please make appropriate changes if you have the time :+1:

PR #6588, opened 2017-10-30
Source decay patch 1

source-decay:
May have been an issue building due to folder name containg upper case chars. Trying with lowercase.


source-decay:
Tried to submit PR and am unsure the Travis CI is saying is wrong. Some sort of issue with creating the Nav

PR #6603, opened 2017-10-30
react native class vs functional components

gwenf:
add a new section to the react native guide.


QuincyLarson:
@gwenf Thanks for writing these articles, and thanks for your patience. These look great! I'm merging them.

PR #6611, opened 2017-10-30
updating framer page

talandaw:
Noticed it was a stub and added information gathered from the Framer website as a basic details starting point

**Description**


QuincyLarson:
@talandaw Thanks for your patience! Thanks for creating this Framer article. This should be super helpful for the community when they're considering design tools like this.

PR #6695, opened 2017-10-30
Added about Threads

KarteekM:
**Description**


KarteekM:
This is my first commit.I will be adding more documentation in the following commit about Threads.

PR #6944, opened 2017-10-30
Ruby: Add Swapcase method to string methos

richsalvucci:
**Description**

I am adding the ruby string method of swapcase to the index.md string method file.


imcodingideas:
@richsalvucci thanks for giving us an example on the `'Hello'.swapcase` method on ruby string documentation. 

This looks good to me. I'm going to merge this in.

PR #6946, opened 2017-10-30
The first meaning

novvladig:
**Description**


beaucarnes:
Thanks for the contribution but this is not the Moscow this article is referring to.

PR #6951, opened 2017-10-30
Creating a Binary Search Tree in Portuguese

krevi27:
To help those who just speak portuguese!

**Description**


fvcproductions:
The ability to translate this content into other languages is currently on the roadmap - see https://github.com/freeCodeCamp/guides/issues/310. 🚗

For the time being, I will close your PR but mark it with a `translation` label so that it may be re-opened in the future after this feature is actually implemented. 🌎

PR #6955, opened 2017-10-30
Working In Tech: Fixing book links

richsalvucci:
The book links to Amazon were not wrapped in anchor tags so a visitor would have to copy and paste the link into a new tab or window. I wrapped the word Amazon so the user can click the word and be taken to the Amazon page for the book in a new window.

The book links to Amazon were not wrapped in anchor tags so a visitor would have to copy and paste the link into a new tab or window. I wrapped the word Amazon so the user can click the word and be taken to the Amazon page for the book in a new window.

### DIFF
--- a/src/pages/working-in-tech/index.md
+++ b/src/pages/working-in-tech/index.md
@@ -15,12 +15,12 @@ Recently Timo Koola provided 10 books every developer should read.
 
 The following four are relevant in all walks of life:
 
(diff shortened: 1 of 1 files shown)


domwakeling:
Hey @richsalvucci

It seems that similar changes have already been accepted earlier for this article you're editing, sorry about that. 😓

If you feel you have more to add, please feel free to open up a new PR.

Thanks again! 😊

---

> If you have any questions, feel free to reach out by commenting below. 💬

PR #6960, opened 2017-10-30
Added clarification

bayem:
**Description**


beaucarnes:
@bayem Thanks for clarifying this.

PR #6970, opened 2017-10-30
Added a benefit of using a foreach loop

Plomboe:
**Description**


fvcproductions:
Thanks so much for contributing changes to the `guides` repo! 🎉

---

> For future reference, it's good practice on GitHub to write add a useful and brief title and description of your changes when creating a PR. 📝

PR #7010, opened 2017-10-30
replaces array-prototype-splice stub with article

ArielLeslie:
Initial creation of article for Array.prototype.splice()

**Description**


src/pages/javascript/standard-objects/array/array-prototype-splice/index.md
```diff
@@ -3,13 +3,59 @@ title: Array.prototype.splice
...
+exampleArray.splice(-1);
+// exampleArray is now ['first', 'second', 'third'];
+```
+
+#### splice(start, deleteCount)
+If a second parameter is included, then `splice(start, deleteCount)` will remove `deleteCount` elements from the array, beginning with `start`.
+```js
+let exampleArray = ['first', 'second', 'third', 'fourth'];
+exampleArray.splice(-1, 2);
+// exampleArray is now ['first', 'fourth'];
```
ezioda004: For this example, the output is `["first", "second", "third"]` because `2` is greater than `arr.length - start` where start is `3` so its removing till `arr.length`, its following this rule:
>if its value is larger than array.length - start (that is, if it is greater than the number of elements left in the array, starting at start), **then all of the elements from start through the end of the array will be deleted.**

ArielLeslie: You're right. I had intended the example to be `splice(1, 2)` rather than `splice(-1, 2)`. Good catch.

ezioda004:
Great write up and thanks for your contribution @ArielLeslie 
Also, congratulations on your first PR merge to this repo! 🎉

PR #7021, opened 2017-10-30
Add methods stub

jlundqui:
**Description**


jlundqui:
Create Go Methods stub

PR #7046, opened 2017-10-30
changed text links into markdown links in the working in tech article 

M0nica:
The links in the working in tech article were written in plain text and not clickable. I updated so that they are formatted properly in Markdown and can be clicked.


beaucarnes:
@M0nica Thanks for your contribution! Sorry for the delay in merging.

PR #7052, opened 2017-10-31
Added back-ticks to properly format code block in the mailto link article

M0nica:
The HTML examples were rendering in the mailto article  as HTML instead of showing the actual HTML code. Updated so that viewers can see the HTML code in the HTML examples (as intended) instead of seeing how the HTML rendered.


beaucarnes:
@M0nica Thanks for catching this and fixing!

beaucarnes:
Just realized that this was already fixed in another pull request. Feel free to open another pull request if needed. 

Many thanks and Happy Coding :+1:

PR #7065, opened 2017-10-31
Add/category progressive web apps

SiriusScaper:
Added a category for progressive web apps and started the article.


SiriusScaper:
Figured out what I did wrong and redid it with proper stub articles for two subsections, app shell, and service workers.

PR #7111, opened 2017-10-31
Added the id attribute

vtisnado:


nathanhannig:
id is a global HTML attribute and not specific to iframes. I don't think it belong is the article.

https://www.w3schools.com/TAGS/att_global_id.asp

https://www.w3schools.com/Tags/tag_iframe.asp

PR #7119, opened 2017-10-31
Add examples of Boolean logic to article on tautologies

strawn-ed:
**Description**
Add an example of a tautology and an example of a contradiction to the article of "Tautologies" in the Logic section.


QuincyLarson:
@strawn-ed Thanks for your patience. Thanks for adding these details to the article.

PR #7125, opened 2017-10-31
src/pages/mathematics/subtracting-fractions/index.md

Reshma205:
Subtraction of Fractions is explained above. Subtraction with denominators same is explained in one example and subtraction with different denominators is explained in another example.

**Description**


Reshma205:
Subtraction of Fractions is explained . Subtraction with denominators same is explained in one example and subtraction with different denominators is explained in another example.

PR #7128, opened 2017-10-31
Updated API

Invince11:
**Description**


nathanhannig:
looks to be copy and paste from https://www.quora.com/What-are-the-advantages-and-disadvantages-of-Application-Programming-Interfaces-api

PR #7191, opened 2017-10-31
add the text "my changes." to article

shivam1512:
Calculating 3 by 3 determinant is as same as calculating 2 by 2 determinant only at a bigger scale.

**Description**


shivam1512:
Calculating 3 by 3 determinant is as same as calculating 2 by 2 determinant only at a bigger scale.

PR #7244, opened 2017-10-31
Remove normalise from preCommit

### LINKED ISSUE #7238: Remove normalise from pre-commit script
We use a pre-commit script that can be found in the `package.json`

We need to remove the normalise part of the pre-commit to stop confusion for contributors when multiple files change when committing.

A future issue will deal with automating the normalise script.

Bob-Thomas:
Removed normalise from the precommit script in package.json.
Closes #7238

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,7 +59,6 @@
   },
   "pre-commit": [
     "pre-commit-message",
-    "lint",
-    "normalise"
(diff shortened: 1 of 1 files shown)


Bouncey:
Thanks @Bob-Thomas, that was quick!

PR #7247, opened 2017-10-31
Added general info.

thee1:
**Description**

PR #7249, opened 2017-10-31
Added general info.

thee1:
**Description**


Bouncey:
As per https://github.com/freeCodeCamp/guides/pull/7247#issuecomment-340716283

PR #7351, opened 2017-10-31
add changes"To select elements with a specific class, we use a (.) na…

rakhi20:
…med as period character, with the name of the class. For example .center {     text-align: center;     color: red; } Here , all HTML elements with class="center" will be red and center-aligned"

**Description**


QuincyLarson:
@rakhi20 Thanks for your patience! Looks good! Merged!

PR #7361, opened 2017-10-31
Added built-in methods descriptions to array type

daunera:
**Description**
I added the three most used typescript built-in methods for array type
Pop, push and reverse.

I hope, you can use it.
Thanks,
Augustine


ryanwhocodes:
@daunera These additions look great, thanks for contributing!

PR #7379, opened 2017-10-31
Add built-in methods description

daunera:
**Description**
I added the three most used typescript built-in methods for string type
split, substr and substring.

I hope, you can use it.
Thanks,
Augustine


joshalling:
This looks good. Thanks @daunera! :+1: :taco:

PR #7390, opened 2017-10-31
Add some built-in features description about typescript tuple

daunera:
**Description**
I added the three most used typescript built-in methods and properties for tuple  type

I hope, you can use it.
Thanks,
Augustine


joshalling:
@daunera Thanks for your contribution! :+1: :+1:

PR #7404, opened 2017-10-31
add changes"In the 'If-else' statement, If the statement within the p…

rakhi20:
…arenthesis is true, the code within the brackets of the 'if' statement will be executed and if the statement within the parenthesis is false, the code within the brackets of the 'else' statement will be executed."

**Description**


joshalling:
Thanks @rakhi20! And congrats on your first contribution! :+1: :volcano:

PR #7406, opened 2017-10-31
area of a parallelogram explained

rohith5890:
Area = bh
area of a parallelogram is base times height ,which is similar to that of a rectangle.

As every parallelogram can be made rectangle, so we use the same formula for rectangle and parallelogram.

**Description**


rohith5890:
Area = bh
area of a parallelogram is base times height ,which is similar to that of a rectangle.

As every parallelogram can be made rectangle, so we use the same formula for rectangle and parallelogram.

PR #7415, opened 2017-10-31
add changes"  Final can be: (1) variable : If we make any variable as…

rakhi20:
… final, we cannot change the value of final variable. (2) method : If we make any method as final, we cannot override it. (3) class : If we make any class as final, we cannot extend the class. "

**Description**


joshalling:
@rakhi20 I'm closing this because it looks like much of this information was already merged in another PR. If you still want to contribute to this stub, rebase your fork and submit another PR with some additional content.

PR #7422, opened 2017-10-31
area of a ellipse explained

rohith5890:
src/pages/mathematics/area-of-a-ellipse/index.md

The area of an ellipse can be found by the following formula area = pi*a*b 
where b is the distance from the center to a co-vertex
a is the distance from the center to a vertex.

**Description**


rohith5890:
The area of an ellipse can be found by the following formula area = pi*a*b 
where b is the distance from the center to a co-vertex
a is the distance from the center to a vertex.

PR #7427, opened 2017-10-31
added an example of while loop

snehaMittal:
**Description**


joshalling:
@snehaMittal Thanks for your contribution! :+1: :crocodile:

PR #7443, opened 2017-10-31
final keyword as class and final keyword as method is added

snehaMittal:
**Description**


joshalling:
@snehaMittal I fixed some markdown issues in your code. Once this passes checks I'll merge.

PR #7485, opened 2017-10-31
Add code in C++ with explanation

abbas10m:
updated article to include  implementation of simple linked list in C++ with explanation

**Description**


abbas10m:
Hi @luandy64 
Thanks for the feedback
I've remove `<conio.h>` as it wasn't required.
As per my understanding, the default return type of `main()` in cpp is int. Works fine without explicitly declaring. Still, I've updated it to avoid any confusion

domwakeling:
Hi @abbas10m. Congrats on your first pull request (PR)! 🎉

PR #7490, opened 2017-10-31
Call apply bind

vishwa9:
**Description**
Added new topic **call**, **apply** and **bind** in JavaScript


joshalling:
@vishwa9 Thank you for your time in putting this explanation together. I think this information would be a great addition to the guide. 

That being said, I'm closing this because the information belongs in the following subfolder: `javascript/standard-objects/function`. The `apply` and `call` stubs are still empty so we would welcome another PR with this information separated into those stubs

PR #7494, opened 2017-10-31
Fix lamp server article

fvcproductions:
Fixes incorrect title for PR that was merged earlier


fvcproductions:
@Bouncey This will fix the earlier title error.

PR #7497, opened 2017-10-31
Update index.md

philliplaibe:
adding git fetch command to check for changes to master file.

**Description**


texas2010:
it seesm it is topic off.  that page about clone. not fetch. it is different. so i have to close this. you have to make new article about fetch.

philliplaibe:
Whoops, since fetch is a way to check for updates from master file, similar to status, I thought it was a good place to put it. I'll try it as a new article.

texas2010:
clone and fetch is different. btw this article about clone.

PR #7511, opened 2017-10-31
Updated info on jQuery

Vajreswarib:
Updated info on jQuery

**Description**


joshalling:
@Vajreswarib I'm closing this because it looks like much of this information already exists in the various sub-folders under jQuery.

PR #7560, opened 2017-10-31
Creating a stub for a general section on unity UI

Fullmight:
I wrote up an introduction/overview to the unity UI, and my current thought is that the way to go is to create a general interface folder, then an introduction sub-folder, on the assumption that keeping high molecularity like this is good but if not hey let me know.


joshalling:
@Fullmight I'm closing this because it doesn't seem to add anything significant to the current content. Feel free to submit another PR with some additional content if you still want to contribute to this stub.

PR #7570, opened 2017-10-31
Remove "React Native" from page title and add setState details.

sealedHuman:
**Description**
"React Native" in the title seems redundant.
Added some details regarding setState and how one should not use "this.state =" outside of the constructor.


BerkeleyTrue:
Deploy preview ready!

Built with commit 1db6efe324d148bc89ed6be832a5aa4e74e88e7f

https://deploy-preview-7570--guide.netlify.com

PR #7572, opened 2017-10-31
Added an example of boosting a random forest.

rk635:
Example of adaboosting and stochastic gradient boost on a random forest. Implementation in python. Dataset: https://www.kaggle.com/c/flavours-of-physics/data

**Description**


joshalling:
Thanks for this contribution @rk635! :+1: :medal_military:

PR #7574, opened 2017-10-31
Added information on k-means clustering

rk635:
Added a description and an example in python on performing k-means clustering on iris dataset.

**Description**


joshalling:
Thanks for your contribution @rk635! :+1: :boar:

PR #7575, opened 2017-10-31
Declaring two classes to the same element css

pimatco:
**Description**


joshalling:
@pimatco Thanks for your contribution! :+1: :+1:

PR #7592, opened 2017-10-31
Added article for Layouts

ladyinblack:

### DIFF
--- a/src/pages/html/layouts/index.md
+++ b/src/pages/html/layouts/index.md
@@ -1,15 +1,20 @@
 ---
 title: Layouts
 ---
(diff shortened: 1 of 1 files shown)


jp-sauve:
Flexbox and floats are two ways to manage layout, but I don't think CSS frameworks that use floats and flexbox can be put in the same category. They aren't a new method, just an easier way of using the previous methods.

joshalling:
@ladyinblack I'm closing this because it looks like much of this information was was already merged in another PR. If you still want to contribute to this stub, rebase your fork and submit another PR with some additional content.

PR #7597, opened 2017-10-31
JS / Standard Objects / String updated index.md

ALeilands:
**Description**

Added info for String standard objects


joshalling:
@ALeilands Thanks for your contribution! :+1: :cat:

PR #7614, opened 2017-10-31
Added null-coalescing operator.

travisterrell:
Added description and example of the null-coalescing operator to the null-conditional page. (The page may need to be re-titled to reflect this.)

**Description**


joshalling:
@travisterrell Thanks for this contribution! :+1: :taco:

PR #7623, opened 2017-10-31
Added Information about WAI, ARIA, and WAI-ARIA

MFry:
The acronyms and technology was mentioned in the basics, but never defined nor formally introduced.

**Description**


amandeepmittal:
Thanks for your contribution @MFry 👍

PR #7648, opened 2017-10-31
Added some clickable indicators

phoolto:
Made it more clear for the user that something is actually clickable.


joshalling:
@phoolto I'm closing this because it looks like a solution for this problem has already been put in place.

PR #7667, opened 2017-10-31
Update index.md

Kyledesma:
updated info regarding R

**Description**


texas2010:
that don't need to change. I am fine with the current article but thank you for the offer.

PR #7669, opened 2017-10-31
Updated Clojure/Conditionals adding info on `cond` and special forms

nmbrgts:
`cond` is a common conditional operator so I added it to the Alternatives to If section. 

I also added a section on special forms. Conditional expressions are a good place to introduce the concept of special forms. Their evaluation rules are intuitive for learners to grasp, but still clearly contrast with Clojure/Lisp's normal evaluation rules. 

I also fixed a small grammatical error.


imcodingideas:
@nmbrgts this is a great pull request. You've added some real value to the `Clojure` conditionals stub. Thank you.

PR #7701, opened 2017-11-01
add Web Components

SiriusScaper:
Added web components and two subsections and started the main article for web components

Adding a web components section as it is another interesting technology gaining traction in recent years. I also started the main article on web components and added two subsections with stub articles.


Samuel-Clarke123:
Closed as this is a beta to something that was merged.

PR #7726, opened 2017-11-01
updated switch description

getUserID:
updated switch description and added source for reference

**Description**


amandeepmittal:
Thanks for your contribution 👍.

PR #7730, opened 2017-11-01
added to description of while loop

getUserID:
added to description of while loop and added source

**Description**


amandeepmittal:
Thanks for your contribution @getUserID!

PR #7751, opened 2017-11-01
Git: add article on Git Hooks

strawn-ed:
**Description**
There is a lot more that should be added to this, but I think hook scripts are an incredibly helpful and powerful tool that would benefit from documentation in this guide.


amandeepmittal:
Thank you for your contribution @strawn-ed 👍

PR #7779, opened 2017-11-01
Add lambda expression example

petrenk0n:
**Description**

### DIFF
--- a/src/pages/python/lambda-expressions/index.md
+++ b/src/pages/python/lambda-expressions/index.md
@@ -17,5 +17,5 @@ print(square(2))
 ### Lambda Expression
 ```python
 square = lambda num: num**2
-print(square(10))
+print(square(2))
 ```
--- a/src/pages/python/lambda-expressions/index.md
+++ b/src/pages/python/lambda-expressions/index.md
@@ -3,10 +3,19 @@ title: Lambda Expressions
 ---
 ## Lambda Expressions
 
+Lambda expression is a breakdown of a function.
+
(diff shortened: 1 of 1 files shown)


src/pages/python/lambda-expressions/index.md
```diff
@@ -3,10 +3,19 @@ title: Lambda Expressions
 ---
 ## Lambda Expressions
 
-Lambda Expressions are ideally used when we have something simple to be done an we are more interested in quickly getting the job done rather than formally naming the function.
-<a href='https://github.com/freecodecamp/guides/tree/master/src/pages/python/lambda-expressions/index.md' target='_blank' rel='nofollow'>Help our community expand it</a>.
+Lambda expression is a breakdown of a function.
```
jp-sauve: I'm not sure what you mean by "a breakdown" of a function. Can you explain what you mean?

petrenk0n: What I mean by that is that lambda expression acts as like an anonymous function. When transforming regular functions into lambda expressions, we simplify the function (deleting some keywords from the function or breaking down the function) by deleting the keyword def and the return statement.

Maybe I should change it to "Lambda expression is an anonymous function"

jp-sauve: That seems to be clearer, but even that assumes you know what an anonymous function is.

petrenk0n: How about this "Lambda expression is an anonymous function - a function that doesn't have a name"

PR #7785, opened 2017-11-01
Update index.md

saradasutar:
**Description**


texas2010:
I don't see a reason to change it. but I am fine with the current. sorry! thank you for the offer to change it.

PR #7848, opened 2017-11-12
Symmetric Difference: Change advanced code solution

ashenm:
* Refined *Problem Explanation* and *Hints*
* Replaced advanced code solution with a snippet that composite of language components that are covered in tutorials to less barricade beginners.
* Replaced documentation references to [MDN JavaScript reference](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference) for parity.
* Fixed old syntax errors


joshalling:
I'm closing this as it looks like this page was removed in a previous merge.

PR #7886, opened 2017-11-24
Regex: corrected regex in utils/regEx.js

chotuaashiq:
#7884 
I Corrected the regex expression in utils/regEx.js as was suggested.
@systimotic please review

PR #7933, opened 2017-12-09
Corrected regex in utils/regEx.js

RohanGurjar:
@systimotic Creating PR as discussed in another PR https://github.com/freeCodeCamp/guides/pull/7886#issuecomment-350201277
1. Changed frontmatter regex.
2. Removed extra spaces from converting-units-minutes-to-hours/index.md file.
3. Updated error message when there's exception while performing regex match.

Please let me know if some more changes is required.

### DIFF
--- a/bin/createNavData.js
+++ b/bin/createNavData.js
@@ -11,8 +11,7 @@ function getPageTitle(content, path) {
   try {
     return content.match(metaTitleRE)[1].trim();
   } catch (err) {
-    throw `Error reading the frontmatter from "${path}/index.md"
-          due to exception ${err}`;
+    throw `Error reading the frontmatter from "${path}/index.md"`;
   }
 }
 
--- a/bin/createNavData.js
+++ b/bin/createNavData.js
@@ -11,7 +11,8 @@ function getPageTitle(content, path) {
   try {
     return content.match(metaTitleRE)[1].trim();
   } catch (err) {
-    throw `Error reading the frontmatter from "${path}/index.md"`;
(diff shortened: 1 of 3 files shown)


BerkeleyTrue:
Deploy preview ready!

Built with commit 45fbf7f52c07a3e2f41cc86c87da264aa58d61ff

https://deploy-preview-7933--guide.netlify.com

bin/createNavData.js
```diff
@@ -11,7 +11,8 @@ function getPageTitle(content, path) {
   try {
     return content.match(metaTitleRE)[1].trim();
   } catch (err) {
-    throw `Error reading the frontmatter from "${path}/index.md"`;
+    throw `Error reading the frontmatter from "${path}/index.md"
+          due to exception ${err}`;
```
systimotic: Thanks for fixing this @RohanGurjar, I apologize for waiting so long to review. This looks great, apart from this line.

Generally speaking, showing the exact error is helpful, but it can be confusing in this case. The person seeing the error likely only wrote articles and didn't touch any code, so an error that `[1] is undefined` (something like that, I'm not sure what the exact error would be 😄) is more confusing and less helpful than just the first line ("error reading frontmatter").

Apart from that, I think this is good to go! Nice work 💯

RohanGurjar: @systimotic I will revert the error message change and update the PR. Thank you :+1:

RohanGurjar:
@systimotic I have made the requested changes. Thank you :+1:

PR #7999, opened 2018-01-13
PHP - Fixed articles with naming issues and removed duplicate articles

nsuchy:
As the title says above this pull request fixes articles with naming issues, it also removes a few duplicate articles.
---


QuincyLarson:
@nsuchy I understand why you're changing the names of some of the articles, but why are you deleting these? Even though these may seem redundant, I can imagine campers searching for these topics. And as far as I can tell, there are no articles with the exact same contents as these articles you're proposing deleting.

Issue #7825, opened 2017-11-06
Articles imported from forum contain invalid nav links


johnkennedy9147:
Looks like some articles which were imported from FCC Forum contain nav links that make no sense within guides.

No idea how many instances there would be.

Example:
![image](https://user-images.githubusercontent.com/26827682/32468409-0e190fcc-c346-11e7-8116-6f69cb1b8c2c.png)

Not sure if there is a quick way to identify affected articles, likewise if can be resolved any way other than manual editing.

thoughts?
---

> ✅️ By submitting this issue, I have verified the following

- [x] [Checked](https://github.com/freeCodeCamp/guides/issues?q=is%3Aissue+is%3Aclosed) to see if the issue has already been discussed before. 🤔️
- [x] *If* proposing new content to be added, made sure enough details were provided. 🔍️

systimotic:
Good catch @johnkennedy9147! I looked into this before, and this was the only instance of this that I could find. Regardless, https://github.com/freeCodeCamp/guides/pull/739 will fix issues like this one where there are three or more links on a single line.

Issue #752, opened 2017-10-06
Pre-commit normalisation changes files


dhanushuUzumaki:
While committing, the pre-commit script changes a lot existing, untouched files and leaves them unstaged.

<img width="1280" alt="screen shot 2017-10-06 at 9 16 49 pm" src="https://user-images.githubusercontent.com/14927280/31286416-b2007216-aadb-11e7-870d-48447048fa2a.png">

Bouncey:
@dhanushuUzumaki thanks for raising an issue about this, believe it or not, this was intended behaviour,  it I don't think it is working out as I imagined. It just causes confusion. 

@systimotic Perhaps we should remove normalise from the pre-commit script and instead have a version of camperbot normalise every day and commit changes if necessary? My experiments with Travis doing it for every PR wouldn't have worked, Travis would have been throwing conflicts everywhere, editing the same files on multiple PRs.

dhanushuUzumaki:
@Bouncey Thanks for letting me know. And once the discussion is complete, if possible I would like to give this a try.

arcdev1:
@Bouncey... for contributors, do you want us to check-in these changes prior to submitting a PR?

systimotic:
@Bouncey After https://github.com/freeCodeCamp/guides/pull/805, I'm fairly confident I could create a plugin that would remove the need for the link normalisation. 
The stub normalisation should only need to happen in certain cases. We could document this and make sure all reviewers are aware of when it needs to be run. 
The linter will already warn if a stub is missing, we'll need to document moving stubs and not being allowed to change stubs if you leave the "This is a stub" text.

systimotic:
Update: I won't have be able to make time to create that plugin any time soon. It also turned out to be a lot more complicated than the other one 😅 I don't feel like I've explored all options yet, but I won't be able to do that soon. Feel free to move ahead as you consider appropriate.

PR #8014, opened 2018-01-25
Fix URL in git hosting stub article

thomaskuntzz:
https://github.com/freeCodeCamp/guides/commit/f7a4426ecd57b790a887b61f9f58a006034a0cb2#r27097759

### DIFF
--- a/src/pages/git/git-hosting/index.md
+++ b/src/pages/git/git-hosting/index.md
@@ -3,7 +3,7 @@ title: Git hosting
 ---
 ## Git hosting
 
-This is a stub. <a href='https://github.com/freecodecamp/guides/tree/master/src/pages/git/tagging-in-git/index.md' target='_blank' rel='nofollow'>Help our community expand it</a>.
+This is a stub. <a href='https://github.com/freeCodeCamp/guides/blob/master/src/pages/git/git-hosting/index.md' target='_blank' rel='nofollow'>Help our community expand it</a>.
 
 <a href='https://github.com/freecodecamp/guides/blob/master/README.md' target='_blank' rel='nofollow'>This quick style guide will help ensure your pull request gets accepted</a>.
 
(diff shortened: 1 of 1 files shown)


src/pages/git/git-hosting/index.md
```diff
@@ -3,7 +3,7 @@ title: Git hosting
 ---
 ## Git hosting
 
-This is a stub. <a href='https://github.com/freecodecamp/guides/tree/master/src/pages/git/tagging-in-git/index.md' target='_blank' rel='nofollow'>Help our community expand it</a>.
+This is a stub. <a href='https://github.com/freeCodeCamp/guides/blob/master/src/pages/git/git-hosting/index.md' target='_blank' rel='nofollow'>Help our community expand it</a>.
```
systimotic: These stubs are usually generated by the `bin/normaliseArticles.js` script. It uses the following template: `https://github.com/freecodecamp/guides/tree/master/src/pages/${filePath}/index.md`. The correct link here would be `https://github.com/freecodecamp/guides/tree/master/src/pages/git/git-hosting/index.md`. Any other link would be changed the next time that script runs.

thomaskuntzz: Yes, still the wrong link... How should a stub be created then if there's a script?

systimotic: @thomaskuntzz `bin/normaliseArticles.js` is the script that creates stubs, but it also does some other things that could mess with the repo right now 😅 We're working on cleaning it up (#7825 #752)

thomaskuntzz:
Cool, thanks for merging!

However, I'm still unsure about how a stub should be created... You mentioned a script or something, what is the preferred way to created a stub?

systimotic:
To follow up on https://github.com/freeCodeCamp/guides/pull/8014#discussion_r164729483, ideally we'd use that existing script, but it's a bit broken right now, so doing it manually is alright. I'm hoping we can fix that script, move the link normalization to a plugin, and maybe put the stub creation in a separate script. The crazy errors I'm getting on https://github.com/freeCodeCamp/guides/pull/739 aren't helping with this 😄

PR #7954, opened 2017-12-20
Fix absolute/relative paths in results

Bouncey:
Closes #7953

PR #8015, opened 2018-01-25
Upgarde all packages to latest, including gatsby

Bouncey:
Upgrade all packages!


Bouncey:
CI fail is in a whitespace error in a file untouched by this PR

systimotic:
New Gatsby still doesn't work in Chrome up to version 63 😢 (https://github.com/gatsbyjs/gatsby/issues/1979). It does work in 64 🙌, but it'll be a couple of months before users catch up to that version (it was released yesterday).

Search is broken, but I'm pretty sure https://github.com/freeCodeCamp/guides/pull/7954 will fix that.

PR #8040, opened 2018-02-01
Write article about Fractions in Maths

Nythiennzo:

---

- [] Added descriptive name to PR
- [] Reviewed necessary formatting guidelines in [`CONTRIBUTING.md`](https://github.com/freeCodeCamp/guides/blob/master/CONTRIBUTING.md).
- [] No plagiarized, duplicate or repetitive  content that has been directly copied from another source.

### DIFF
--- a/src/pages/mathematics/fractions/index.md
+++ b/src/pages/mathematics/fractions/index.md
@@ -4,13 +4,13 @@ title: Fractions
 
 ## Fractions
 
-A fraction represents part of a whole number, a value between 0 and 1.
+A fractional number represents part of a whole number, a value between 0 and 1. They can be represented using decimals (e.g. 0.1), percentages (e.g. 10%), negative indices (e.g. 10 ^ -1) or fractions (e.g. 1/10).
 
-A fraction normally has two parts:
+A fraction normally has two parts which are divided by a line:
 
-* Numerator: Represents the number of equal parts marking up the fraction.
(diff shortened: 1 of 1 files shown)


thomaskuntzz:
Hi @Nythiennzo, thanks for your contribution! 👌

In your article, you say that a fraction has a value between 0 and 1, which is true, but I think it would be good to mention the difference between a fraction and a fractional number. Would you feel like adding that? 😉

Nythiennzo:
I'm doing it right now!

Nythiennzo:
It's done.
Please check it @thomaskuntzz

alinepegas:
Hi @Nythiennzo.
What do you think about mentioning a fraction is a representation of a rational number?
Take a look at wikipedia:
[Rational Number](https://en.wikipedia.org/wiki/Rational_number)

luandy64:
Hey @Nythiennzo 

Maybe this is just me, but the sentence you wrote, "A fractional number represents part of a whole number, a value between 0 and 1.", makes it sound like you are defining a "whole number" to be "a value between 0 and 1". Can we change that?

Nythiennzo:
Hi @alinepegas,

Thanks for the suggestion.  I got the understanding that a rational number is a number that can be expressed as the fraction p/q where:
- p and q are integers
- q is not 0
 However this does not mean that only rational numbers are fractions.

For example e/10 is not a rational number but would be considered a fraction

For this reason, I do not think mentioning rational numbers is necessary.

PR #8042, opened 2018-02-01
fix #8028 broken breadcrumbs

### LINKED ISSUE #8028: Broken breadcrumbs in search results
#751 added the path of the articles showing up in the search results. However, the breadcrumb with the path doesn't show up right. There's this weird `Https: >  > Guide.freecodecamp.org >` prefix now in all the paths on the search results...

screen shot 2018-01-29 at 3 33 45 pm
screen shot 2018-01-29 at 3 34 29 pm
screen shot 2018-01-29 at 3 34 34 pm

dhanushuUzumaki:
fix for #8028

### DIFF
--- a/src/pageComponents/search/Results.jsx
+++ b/src/pageComponents/search/Results.jsx
@@ -83,7 +83,9 @@ class Results extends PureComponent {
             </Media.Left>
             <Media.Body>
               <Media.Heading>{title}</Media.Heading>
-              {_index === 'guides' && <Breadcrumbs path={url} />}
+              {_index === 'guides' &&
+                <Breadcrumbs path={url.replace(isGuidesUrl, '')} />
+              }
               <p>{description}</p>
             </Media.Body>
           </Media>


thomaskuntzz:
Seems to do the trick, thanks a lot!

The only thing I see that could be improved is that the excerpt text and article title are still underlined when hovering only the folder in which the article is:

screen

systimotic:
@thomaskuntzz Good catch! This made sense on the article page, because it would be weird to have a link to the current page, but an underline would be good on the search page (I think the color should stay gray). Do you want to open an issue (and maybe a PR too) about this?

thomaskuntzz:
@systimotic I was talking about the article title and excerpt being underlined even when I'm only hovering the breadcrumb, not about the link of the article itself being grey in the breadcrumb. But as each search result item is wrapped in one link, it'll be difficult to underline only the breadcrumb when I'm hovering it, since when hovering the breadcrumb I'm also hovering the wrapper link which leads to the article...

And something else: why do the search results vanish when the search bar blurs? Is it the expected behavior? I find it really annoying, and as a bonus is makes the search results impossible to inspect in the browser...

thomaskuntzz:
I opened #8045 for the search results vanishing when the search bar blurs. I think we can merge this. @Bouncey Can you review this please? 🙂

systimotic:
Stuart is busy trying to get the freeCodeCamp beta live, which is probably why he hasn't been responding here. I can confirm that this fixes the issue we were seeing, so I'll merge it. If it needs improvements, @Bouncey is welcome to let us know when he has time 😃

Thanks @dhanushuUzumaki and @thomaskuntzz!

PR #8091, opened 2018-02-26
fix(ENOSPC): Increase inotify watchers during build

Bouncey:


Bouncey:
I can build this locally using the build image.

Bouncey:
There must be an issue with concurrent builds at netlify

PR #8141, opened 2018-04-27
Use proper em dashes in README

thomaskuntzz:
The proper dashes to use are em dashes when the goal is to indicate a break in the sentence

https://en.wikipedia.org/wiki/Dash

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@
 
 The freeCodeCamp community is building a massive, searchable **"Guide"**.
 
-This reference tool will eventually include thousands of articles that cover all areas of development, design, and data science - all written to be easily understood by people new to coding.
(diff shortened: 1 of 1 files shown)


thomaskuntzz:
@pertrai1 That's the typographically correct way to use em dashes in this case. I just checked and in fact both en dashes with spaces and em dashes without spaces can be used.

thomaskuntzz:
The most important thing is consistency, so if en dashed with spaces are used everywhere this PR might not be relevant

erictleung:
I was worried that they would render incorrectly or confusingly, but they turned out fine.

image

amandeepmittal:
@thomaskuntzz Thanks for your contribution 👍.

PR #8242, opened 2018-07-08
Get a hint - RWD: Introduction to HTML5 elements

R-Layer:
Stub removed.
This hint focuses more on the meaning of 'wrap elements' rather than on the 'new' tags introduced by the HTML5  or the semantic HTML itself: that's because that is the 'real' difficulty of the challenge judging by the posts on the forum.

---


imcodingideas:
@R-Layer it's me again! Thank you for taking the time to write this stub. You're amazing.

R-Layer:
Ah i just noticed the merges from the mail notifications ^^ 

Once for all, thanks for the good words and for the work you all do every day! We all owe you!

PR #8262, opened 2018-07-12
Get a hint - RWD: Create an ordered list

R-Layer:
Stub removed.
Remarked the differences between `ul` and `ol` ( or better, remarked that there are almost no difference but the tag ) and given an example.
Provided a link to the previous challenge ( since the list 'pattern' is the same if something is not clear it could be referred to the previous challenge aswell ).

---


R-Layer:
Not sure what's wrong or how to fix.
I think is the link to gives problem but how to change it? Does it need another branch just for that?

PR #8290, opened 2018-07-21
Delete hello-word.md

imcodingideas:
This is fixing a merge request that's breaking some builds.


imcodingideas:
This looked good to me when I merged it but; because this was old we have what looks like a newer structure for the .md files.

With that said I need to merge this hotfix in.

PR #8311, opened 2018-07-25
Create a Stub for Puppet

JohnSprunger:


imcodingideas:
@JohnSprunger Thanks for taking the time for creating the stub template for the Puppet.

PR #8313, opened 2018-07-25
Create a Stub for XML

JohnSprunger:
Extensible Markup Language


imcodingideas:
@JohnSprunger this looks great. By creating this stub request it will give someone the opportunity to write about XML.

PR #8418, opened 2018-08-24
Fix markdown error in PR template

divayprakash:
Correctly renders checkboxes which were earlier being rendered as a list with brackets

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -14,12 +14,12 @@ Before creating a PR, please make sure to verify the following by marking the ch
 
 ## ✅️ By submitting this PR, I have verified the following
 
-- [] Added descriptive name to PR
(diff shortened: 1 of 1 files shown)


danhmanh:
[x ]

danhmanh:
I think its not neccessary because when you add space between 2 bracket like this `[ ]`, participant have to delete a space to make markdown enable. `[x ]` doesnt work.

erictleung:
@danhmanh I think the intent in this PR is that you create the pull request, and you can just click on the checkboxes to change them when you view them. This is probably in bikeshedding territory and really up to educating contributors that there shouldn't be an extra space.

I'll wait for others opinions on the matter. Given my cursory look at PRs, I think adding this space may confuse more people as you point out.

joshalling:
I'll merge this since the freeCodeCamp and curriculum templates have the space. This way we will be consistent across the repos. Thanks @divayprakash!

PR #8419, opened 2018-08-24
Fix markdown error in ISSUES template

divayprakash:
Correctly renders checkboxes which were earlier being rendered as a list with brackets


joshalling:
Thanks for your contribution @divayprakash 👍 👍

PR #8443, opened 2018-08-29
Added some minor changes for beginners

hxw1302:
I just added a minor change to the answer to reflect my thought on this, I use Boolean to determine if the prop exists or not. if the prop is undefined it will return false. I think this will create a better understanding for beginners.

---


joshalling:
@hxw1302 It looks like there is an error in your solution so it doesn't pass the tests in the curriculum. Also, if you want to add another solution to this stub please leave the one that is already there and add a separate one with your code below. Feel free to submit another PR with those changes if you still want to contribute.

PR #8497, opened 2018-09-04
Fix sorted union alternative advanced solution

kr3at0:
Edit the alternative advanced solution the the [Sorted Union](https://learn.freecodecamp.org/javascript-algorithms-and-data-structures/intermediate-algorithm-scripting/sorted-union) challenge in section Javascript Algorithms And Data Structures Certification > Intermediate Algorithm Scripting. Either use explicitly the past arguments through a parameter in the function (which is better, imho) or remove the parameter at all and then use the arguments key word which would create a problem if there needs an additional parameter for the function that specifies depth, for example, or anything else.

---


joshalling:
This looks good! Thanks, @kr3at0! :+1: :ghost:

PR #8507, opened 2018-09-06
Added hint and solution for problem

josephwhittington:
I just added a solution and two simple hints to the last problem in the regex problem set for the javascript data structures and algorithms certification curriculum because I noticed it was missing.


amandeepmittal:
Congratulations on your first contribution @josephwhittington 🎉

PR #8554, opened 2018-09-13
Add article on why we use const for action types

galexynye:
Just a short explanation of the reason why using const over strings is beneficial for Action Types in redux.


joshalling:
Thanks @galexynye and congrats on your first contribution! :+1: :volcano:

PR #8682, opened 2018-09-24
Sample Solution Added

dhpaulwooton:
I just added a sample solution, no explanations


QuincyLarson:
@dhpaulwooton Thanks for your contribution. It looks like other contributors have since added a more comprehensive article.

PR #8864, opened 2018-10-01
Added links to the programs made with Electron, and added Discord to list

15fine:
---
Added links to the programs made with Electron, and added Discord to list
---
Hey! So I added links to the programs listed under the "Programs made by Electron" list, so that anyone reading it can see how useful electron is!
I also added Discord to the list.
-->


joshalling:
Thanks @15fine! :+1: :ghost:

PR #8951, opened 2018-10-01
Added link to great article by Dan Abramov

dominikbulaj:
You Might Not Need Redux, an article by Dan Abramov, creator of Redux, who points out when we really need Redux has been added.


dominikbulaj:
@amandeepmittal adjusted. Thanks for feedback and sorry I missed it on a first try

PR #9208, opened 2018-10-02
Added a Note.

alindsharmasimply:
Added some additional information about an escape sequence. 😊

---

### DIFF
--- a/src/pages/c/hello-world/index.md
+++ b/src/pages/c/hello-world/index.md
@@ -15,7 +15,7 @@ To write on console you can use the function `printf()` contained in the library
      return 0;
  }
  ```
-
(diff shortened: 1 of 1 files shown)


honmanyau:
@alindsharmasimply Thank you very much for your pull request! Unfortunately I have closed it because I feel that it doesn't add substantial amount of information to the example and you have simply stated a fact without explanation.

With the fact that the note is meant to help people with less experience, and with this particular PR in question, perhaps you should explain why an a new line character is a good idea in C, point out what would happen if you don't do that, why is it generally a good idea to include it, and perhaps even expand on why in some cases experienced programmers also include `\n` at the beginning of a string.

PR #9255, opened 2018-10-02
Updated the state file to include more information about State with e…

highskillzz:
Updated the state file to include more information about State with examples and also explained the code related to state as well as how it is used in different places


highskillzz:
I have made the changes. Please check @Asjas

highskillzz:
What changes do you require @Asjas? I don't know if this is the earlier message but it shows you have asked for more changes.

PR #9280, opened 2018-10-02
Made changes to the file to include more about components

highskillzz:
Updated the file to include more about components like differences between stateful and stateless componens with examples as well explained more about components

### DIFF
--- a/src/pages/react/components/index.md
+++ b/src/pages/react/components/index.md
@@ -4,7 +4,79 @@ title: Components
 
 # Components
 
-Components are the building blocks of React.  
+Components are self-sustaining, independent micro-entities that describe a part of your UI. An application's UI can be split up into smaller components where each component has its own code, structure, and API.   
+
+The reason why react is famous can be partly attributed to the components architecture. Basically components are divided parts of a big application. 
+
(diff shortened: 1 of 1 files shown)


ezioda004:
Hi @highskillzz,
Almost all of the article you wrote is taken from [this article](https://code.tutsplus.com/tutorials/stateful-vs-stateless-functional-components-in-react--cms-29541) with no citation to the original source.
I'm sorry we cannot accept these changes. 😞 
Please read our [contributing](https://github.com/freeCodeCamp/guide/blob/master/CONTRIBUTING.md) guide. We do not condone plagiarism.

highskillzz:
I have taken three paragraphs from this article but the rest is unique and would it solve the problem if I put this article under more information?

ezioda004:
@highskillzz I'd highly suggest that you read the [attributions](https://github.com/freeCodeCamp/guide/blob/master/CONTRIBUTING.md#attributions) section first. 
Any material that you're using directly and unchanged, should be quoted and cited.

You can most certainly cite the paragraphs like it shows in the example of the link I posted above.
I hope that cleared your doubt, if you have any more questions, feel free to ask! 😄

PR #9287, opened 2018-10-02
Changed the npm install to local by adding --save and a few other cha…

highskillzz:
Changed the npm install to local by adding --save and removed the number 1 above redux and labelled the nameless function as it would be a bit confusing to beginners

### DIFF
--- a/src/pages/redux/redux-thunk/index.md
+++ b/src/pages/redux/redux-thunk/index.md
@@ -3,7 +3,7 @@ title: Redux Thunk
 ---
 ## Redux Thunk
 
-Redux Thunk is middleware that allows you to return functions, rather than just actions, within Redux<sup>1</sup>. This allows for delayed actions, including working with promises.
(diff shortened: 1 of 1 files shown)


src/pages/redux/redux-thunk/index.md
```diff
@@ -3,7 +3,7 @@ title: Redux Thunk
 ---
 ## Redux Thunk
 
-Redux Thunk is middleware that allows you to return functions, rather than just actions, within Redux<sup>1</sup>. This allows for delayed actions, including working with promises.
+Redux Thunk is middleware that allows you to return functions, rather than just actions, within Redux. This allows for delayed actions, including working with promises.
```
Asjas: Hi @highskillzz 

Please add the change back that you are removing on this line. The source link is at the bottom of the page for this example.

highskillzz: ohh ok

Asjas:
Fixed in #9348

PR #9296, opened 2018-10-02
Fix to use csharp instead of C# in code blocks

kylegalbraith:
Fix to use csharp instead of C# in code blocks. This should resolve the following issue in the guide.

image

- [x] No plagiarized, duplicate or repetitive content that has been directly copied from another source.


paulywill:
@kylegalbraith Good catch!

PR #9351, opened 2018-10-03
Updated the index.md to include more information about docker

highskillzz:
Included more information about docker as well as also referered the sources from which the information was taken and also added learning resources


paulywill:
@highskillzz Content within has been flagged as plagiarism.

PR #9381, opened 2018-10-03
Added additional information about For Loops

NatPirch:
Added additional information about For Loops and their use in sorting algorithms.

---

- [ x] Added descriptive name to PR
- [x ] Reviewed necessary formatting guidelines in [`CONTRIBUTING.md`](https://github.com/freeCodeCamp/guides/blob/master/CONTRIBUTING.md).
- [x ] No plagiarized, duplicate or repetitive  content that has been directly copied from another source.

### DIFF
--- a/src/pages/c/for/index.md
+++ b/src/pages/c/for/index.md
@@ -6,6 +6,8 @@ title: For Loop
 
 The `for` loop executes a block of code until a specified condition is false. Use `while` loops when the number of iterations are variable, otherwise use `for` loops. A common use of `for` loops are array iterations.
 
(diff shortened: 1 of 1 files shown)


honmanyau:
@NatPirch Thank you very much for submitting a pull request! While the sentence you have added is not irrelevant, I feel that it is also not adding a substantial amount of information that improves the understanding of the reader.

I have tagged it as `changes requested` for now in case you would like to improve your edits; here are some suggestions:

* **Why** the `for` loop are often used in sorting algorithms (instead of other constructs).
* **bubble** and **selection** are not actually code and I think there is no need to wrap them with back ticks.

Thank you very much for your help in advance!

PR #9405, opened 2018-10-03
Modified with clear Intro

sem1colon:
Proper Introduction is given.


honmanyau:
@sem1colon Thank you very much for updating the introduction. I will not be merging your PR because it's appears to have been copied verbatim from another source without reference.

PR #9450, opened 2018-10-03
Added note to php sessions

larsdahl:
Added a note about outputs in cookie-based session.


QuincyLarson:
@larsdahl Thanks for adding this important detail.

PR #9545, opened 2018-10-04
Cplusplus:edit for loop article

Artistic18:
The information was not so descriptive.Moreover the in general the three parts of a for loop statement is defined as initialization,condition and update. In the previous description the name of update expression was declared incorrectly as 'increment' which becomes a special case because the update operation can also be applied for decrementation (i.e : a-- operator).Moreover I also added an explanation for the code because an explanation is always needed whenever you are writing a guide.

Better explanation(much more descriptive). Explanation for the code is provided.Easy to understand.
---

  - Your PR should NOT be called `Update index.md`, since it does not help the maintainer understand what has been changed.
  - Additional PR title examples:
    - **Git: edit Git Commit article**
    - **PHP: create PHP section and add Data Structures article**
- [ ] No plagiarized, duplicate, or repetitive content that has been directly copied from another source.

### DIFF
--- a/src/pages/cplusplus/for-loop/index.md
+++ b/src/pages/cplusplus/for-loop/index.md
@@ -3,43 +3,52 @@ title: For Loop
 ---
 
 # For Loop
-A for loop is a repetition control structure that allows you to efficiently write a loop that needs to execute a specific number of times.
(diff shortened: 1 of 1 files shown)


paulywill:
@Artistic18 There were some spelling and conflicts.

Issue #9608, opened 2018-10-05
Suggestion: Remove ES6 folder


ezioda004:
There's an ES6 folder in `/pages/javascript/es6` which have like 4 articles. I propose we flat this folder and move the articles directly to the `javascript` folder because:
1. There's already ES6 features articles added in the JS folder directly, example: arrow function, classes, promises, template literals etc. 
2. JS have yearly releases so since we have ES6 folder, we should add ES7/8/9 folders as well which from maintaining POV seems like unnecessary extra work.  
3. Confusion: There's already duplicate articles (example: arrow functions) because the people who write the articles don't bother looking into nested folder (such as ES6). Live example: https://github.com/freeCodeCamp/guide/pull/9579

Thoughts? 
cc: @Bouncey, @amandeepmittal , @Asjas, @honmanyau

honmanyau:
@ezioda004 I tend to agree, although it may be a good idea to leave it till after the updates are rolled out and guides are merged. :)

honmanyau:
Argh! I accidentally clicked the close button!

joseph-allen:
PR ready if you need it :)

QuincyLarson:
@joseph-allen I think we can leave this folder. It will gradually expand as new features are introduced, and not all of these are yet implemented in all browsers, so it's worth having a distinction. Also keep in mind that a lot of people will be searching specifically for ES6 features, and having this subfolder will make their search easier.

ezioda004:
Hey @QuincyLarson, thanks for clearing it up. I have one question, what do we do if the articles already exists outside the ES6 folder and there's PR to add the same article in ES6 folder? Example [this](https://github.com/freeCodeCamp/guide/blob/master/src/pages/javascript/spread-syntax/index.md) article and this PR https://github.com/freeCodeCamp/guide/pull/9820. Can we merge that PR?

PR #9757, opened 2018-10-06
move content out from ES6 #9608

joseph-allen:
Move content from the es6 folder to prevent future nesting of versions, also we are seeing duplicate content off the back of this.

solves issue #9608 if needed

---

- [ x] Reviewed necessary formatting guidelines in [`CONTRIBUTING.md`](https://github.com/freeCodeCamp/guides/blob/master/CONTRIBUTING.md).
- [ x] No plagiarized, duplicate, or repetitive content that has been directly copied from another source.


QuincyLarson:
@joseph-allen Thanks for your contribution. I commented on the article. I don't think we need to remove folders - we can have a lot of sub folders and that will aid browsing and discoverability.

PR #10020, opened 2018-10-09
Revert "Better "Structured View" of Time Complexity"

paulywill:
Reverts freeCodeCamp/guide#9403


honmanyau:
@paulywill Just a note for after the updates have been rolled out—the suggestion to change the format to O(n<sup>2</sup>) instead of using O(n\*n), other than formatting conventions, is so that there is no more confusion with escaping markdowns.

The new format (combining points into a single line using "respectively") is also less clear in my opinion, and I'm not sure that using a space to "escape" `*`, instead of using `\`, is a considered best practice.

PR #10032, opened 2018-10-09
Correction of alignment in code examples

Nik-Pavlov:
I would like to suggest some small corrections of alignment in code examples

---

## ✅️ By submitting this PR, I have verified the following:

- [ ] Added descriptive name to PR
  - Your PR should NOT be called `Update index.md`, since it does not help the maintainer understand what has been changed.
  - For example, if you create a **Variables** article inside the **Python** directory, the pull request title should be **Python: add Variables article**.
  - Additional PR title examples:
    - **Git: edit Git Commit article**
    - **PHP: create PHP section and add Data Structures article**
- [ ] Reviewed necessary formatting guidelines in [`CONTRIBUTING.md`](https://github.com/freeCodeCamp/guides/blob/master/CONTRIBUTING.md).
- [ ] No plagiarized, duplicate, or repetitive content that has been directly copied from another source.

PR #10034, opened 2018-10-09
C++ syntax highlightning and small naming fix

Nik-Pavlov:
I would like to suggest some small correction in code block in order to highlight c++ syntax.
Also, I think that it will be correct to say `Printing string` instead of `Printing character` because 'character' corresponds to a single symbol, not a sequence of symbols like in the code example. So I changed method input parameter type to ` string` and output text.

---

  - Your PR should NOT be called `Update index.md`, since it does not help the maintainer understand what has been changed.
  - Additional PR title examples:
    - **Git: edit Git Commit article**
    - **PHP: create PHP section and add Data Structures article**
- [ ] No plagiarized, duplicate, or repetitive content that has been directly copied from another source.

### DIFF
--- a/src/pages/cplusplus/overloading/index.md
+++ b/src/pages/cplusplus/overloading/index.md
@@ -13,8 +13,9 @@ You can have multiple definitions for the same function name in the same scope. 
 
 Following is the example where same function print() is being used to print different data types âˆ’
 
-```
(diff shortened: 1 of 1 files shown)


honmanyau:
Closed as per the comment made in #10032.

PR #10060, opened 2018-10-09
React: Pass Props to a Stateless Functional Component

prokawsar:
Added Article about "Pass Props to a Stateless Functional Component"
And Given an example.


ezioda004:
Hi @prokawsar,
Good explanation but there is no solution to the [challenge](https://learn.freecodecamp.org/front-end-libraries/react/pass-props-to-a-stateless-functional-component/) in your article.

prokawsar:
Should I give a direct solution? @ezioda004

ezioda004:
@prokawsar Yes, you could add a Spolier heading and provide a solution to the challenge.

prokawsar:
@ezioda004 Solution added

PR #10143, opened 2018-10-10
React: Render Conditionally from Props

prokawsar:
Simply solution added


joshalling:
Thanks for the contribution @prokawsar! :ghost: I just fixed some typos.

PR #10204, opened 2018-10-10
Give Sibling Elements a Unique Key Attribute

prokawsar:
Added simple explanation and solution


ezioda004:
Thank you @prokawsar for your contribution!

PR #10211, opened 2018-10-10
Get State from the Redux Store

prokawsar:
Added solution


imcodingideas:
@prokawsar thanks showing the solution. This will help millions of people who read the guide!

PR #10215, opened 2018-10-10
Dispatch an Action Event

prokawsar:
Added solution


imcodingideas:
@prokawsar you're on fire, man! Thank you so much for your contribution.

PR #10308, opened 2018-10-11
updating hint for react:introducing inline styles

Kaleb-kougl:
Adding a step by step walkthrough of how to approach JSX attributes and also adding a final solution.


paulywill:
@Kaleb-kougl 

WELL DONE

PR #10333, opened 2018-10-11
Electron: Included Skype as an example

DavSanchez:
Hello! Just a little update here to include Skype in your list of Electron apps. I'm trying to complete this year's Hacktoberfest challenge... Feels like cheating a bit to be honest! 😅


paulywill:
@DavSanchez Thank you!

